PR #26, opened 2014-11-04
Removed redundant $alternate_base_url switching

### LINKED ISSUE #15: Url to combined css and js missing first slash
It says in the readme, that you are supposed to put the static_base_url without trailing slash in the config file.

On our current project I could not get that to work in combination with combined ressources. The url kept missing the "/" between domain and themes dir.

I'm not certain if this is a staticpublishqueue issue or a problem with combining files.

jonom:
The `$alternate_base_url` switching code below doesn't seem to serve a purpose. I'm sure it was probably put there for a reason so please correct me if I'm wrong - but I couldn't find any positive impact it had on the generated cache files. It does however have some negative impacts:

* causes all links in the generated page to become absolute, which is not necessary because of the included base tag, and also makes it inconsistent with the live site
* documentation states that `FilesystemPublisher::$static_base_url` must not include a trailing slash, but omitting this in combination with the absolute link conversion causes all paths to assets to be broken (e.g. mysite.comthemes/css/etc.)
* causes some mangling of links that point to pages on other subsites

Fixes #15

### DIFF
--- a/code/model/FilesystemPublisher.php
+++ b/code/model/FilesystemPublisher.php
@@ -147,7 +147,7 @@ class FilesystemPublisher extends DataExtension {
 
 			if (Config::inst()->get('FilesystemPublisher', 'domain_based_caching')) {
 				if (!$urlParts) continue; // seriously malformed url here...
-				$filename = $urlParts['host'] . '/' . $filename;
+				if (isset($urlParts['host'])) $filename = $urlParts['host'] . '/' . $filename;
 			}
 		
 			$mappedUrls[$url] = ((dirname($filename) == '/') ? '' :  (dirname($filename).'/')).basename($filename);
--- a/code/model/FilesystemPublisher.php
+++ b/code/model/FilesystemPublisher.php
@@ -194,10 +194,6 @@ class FilesystemPublisher extends DataExtension {
 		$currentBaseURL = Director::baseURL();
 		$staticBaseUrl = Config::inst()->get('FilesystemPublisher', 'static_base_url');
 		
-		if($staticBaseUrl) {
-			Config::inst()->update('Director', 'alternate_base_url', $staticBaseUrl);
-		}
-		
 		if($this->fileExtension == 'php') {
 			Config::inst()->update('SSViewer', 'rewrite_hash_links', 'php'); 
 		}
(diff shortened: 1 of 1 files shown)


ss23:
I believe this change breaks the case described by @wilr 

>the correct way to publish to a different domain would be to change the value of `FilesystemPublisher::$static_base_url`... I think?
Since that value isn't used anywhere now, it's a) useless, and b) means you can't publish to a different URL

May we have this reverted until sufficient tests can be completed? Currently the tests are overriding alternate_base_url too, which may be what we decide to do, but as it stands, the documentation implies this isn't required.

For the original issue, I haven't been able to reproduce it yet, but perhaps if you can provide a test case for it specifically @jonom we can see if this is the only fix.

sminnee:
The use-case that this setting is designed for is where you host the CMS server on a different domain from the live server.

So you have `cms.mysite.com` sitting behind a firewall, and `www.mysite.com` available publicly. Static publisher is used to generate HTML files on the CMS server and then once they're generated, rsync them out to the public server.

When those files are rendered, they need to use `www.mysite.com` as the absolute URL of the pages (e.g. for the `<base>` tag), even those the `Director::absoluteBaseURL()` return value will be `cms.mysite.com`. We can't override base URL for the entire CMS instance—that will break other things—which is why `FilesystemPublisher.static_base_url` is available as a config option.

jonom:
Hi @ss23 I apologise if I've inadvertently broken something but I've just tried setting this with this patch in place on a local site:
```yaml
FilesystemPublisher:
  domain_based_caching: true
  static_base_url: http://staticsite.dev
```
And the generated html files (found in cache/test.dev/ directory) included a base tag for `<base href="http://staticsite.dev/">` and relative URLs as expected. I had a few absolute URLs on the page too and these were formed correctly and began with http://staticsite.dev/.

This tells me that the FilesystemPublisher::$static_base_url setting is still being respected with this patch in place, and it should still be possible to export a site to whatever URL you like.

I tried generating the cache files again with `domain_based_caching` set to false and got the same results which I thought was a bit odd but a little digging showed me this is overridden to true if you're using subsites. Perhaps this patch only works as expected when domain_based_caching is set to true?

If I have time later today I will try this patch on a fresh site that doesn't use subsites and see what results I get.

PR #30, opened 2015-01-14
Add ability to skip pages being cached

barry-bookinglive:
When there are pages that we don't want published statically (e.g. pages with access restrictions by group) this allows us to do that with the least change required to the module.

### DIFF
--- a/code/tasks/SiteTreeFullBuildEngine.php
+++ b/code/tasks/SiteTreeFullBuildEngine.php
@@ -99,8 +99,9 @@ class SiteTreeFullBuildEngine extends BuildTask {
 		// Collect all URLs into the queue
 		foreach($pages as $page) {
 			
-			if (is_callable(array($page, 'canCache')) && !$page->canCache())
+			if (is_callable(array($page, 'canCache')) && !$page->canCache()) {
 				continue;
+			}
 				
 			if (is_callable(array($page, 'urlsToCache'))) {
 				$this->getUrlArrayObject()->addUrlsOnBehalf($page->urlsToCache(), $page);
--- a/code/tasks/SiteTreeFullBuildEngine.php
+++ b/code/tasks/SiteTreeFullBuildEngine.php
@@ -102,7 +102,7 @@ class SiteTreeFullBuildEngine extends BuildTask {
 		foreach($pages as $page) {
 
 			if (is_callable(array($page, 'canCache')) && !$page->canCache()) {
-				$arrNonCachedPages[] =
+				$arrNonCachedPages[] = $page;
 				continue;
 			}
 
--- a/code/tasks/SiteTreeFullBuildEngine.php
+++ b/code/tasks/SiteTreeFullBuildEngine.php
@@ -44,7 +44,7 @@ class SiteTreeFullBuildEngine extends BuildTask {
 	}
 
 	/**
-	 * 
+	 *
 	 * @param SS_HTTPRequest $request
 	 * @return bool
 	 */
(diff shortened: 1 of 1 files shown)


mediabeastnz:
?

dhensby:
@wernerkrauss I agree, but in @barry-bookinglive's defence, it is the exact same approach already used in the method. You should justify why it should be done differently

barry-bookinglive:
As discussed braces added

dhensby:
Ok, I fear that this could cause a bug:

the `isCacheable` function can change it's return value based on the data of the Model. This could mean a Page goes from a currently cached state to a state where it can't be cached; does this change mean that the cached file would remain in tact (and thus stale and cached) or would it be removed?

I fear it would not be removed.

barry-bookinglive:
Is there any update on how this would alleviate your fears?
I've just returned here based on two issues both UDF contact forms returning 302 on submission because of this - it's understood that there are some cases that would cause a fault if the value changed... so should we instead focus on...

1) only add constants to this list 
2) hard code the cases - "if ($page->Classname == 'UserDefinedForm')"

how about that?

dhensby:
We probably need a more aggressive way of clearing cache items.

Building a list of all the pages that should have been cached and then removing the ones off the file system that exist but shouldn't.

barry-bookinglive:
How about at the point of checking it in runFrom either add the page to the queue to be cached OR remove it from the cache if we don't' want to cache it?

In addition a task to do just the later could be run often (more often than the full rebuild)

dhensby:
I don't mind how it's achieved, but we need to store a list of URLs that *can't* be cached and we need to make sure the cache files for those are expired (ie: deleted) - we can't just take no action on items that report as being uncachable

barry-bookinglive:
why do we need to store a list?
provided the system cleans them up regularly - or on demand that should be enough right?

If there might be files left over after a full build then how about noting the time at the start, then delete all files that are older than that time at the end of the process?

dhensby:
@barry-bookinglive I just mean a temp list as an array whilst building the cache.

Deleting anything older than the time when the process starts will be tricky because what if you don't cache every file on this run? What if several servers are running different parts of the queue?

barry-bookinglive:
OK how about the latest commit? It's untested - but is it in the right direction?

dhensby:
Yep, that's the kind of thing.

This means that only files that the current execution is responsible for "publishing" (or invalidating) will be considered for removal, which sounds about right.

Are there any instances where a invalidation would need to cascade?

code/tasks/SiteTreeFullBuildEngine.php
```diff
@@ -96,14 +96,33 @@ protected function runFrom($start) {
...
 				$count++;
 			}
 		}
 
+		$publisher = singleton('SiteTree')->getExtensionInstance('FilesystemPublisher');
+		if ($publisher) {
+			foreach ($arrNonCachedPages as $page) {
+				$pathsArray = $publisher->urlsToPaths(array($page->Link()));
+				$filePath = $publisher->getDestDir() . '/' . array_shift($pathsArray);
+				@unlink($filePath);
```
dhensby: Please don't silence errors

barry-bookinglive: Sorry was a cut and pasted from this module... they must have been removed recently

barry-bookinglive:
>>Are there any instances where a invalidation would need to cascade?

Do you mean like a page not being cached would mean child pages wouldn't [anymore] be cached too?

dhensby:
Precisely

barry-bookinglive:
Hmmmm well I'm not sure I see it that way for any of my examples.  So the reporting case here is the the user defined form...  and I'm perfectly happy to have this page not cached - but any children of it cached...   why not?

If this is in another reported case where a page has been restricted by access level... well the same should apply each page can be considered on it's own.  I'm not sure of a case where the parent would be restricted but not the child... but that would be handled correctly by considering each page on it's own, right?

dhensby:
> but that would be handled correctly by considering each page on it's own, right?

I guess so as long as `canCache` returns false when taking into account all these restrictions.

I think we're at a point where this is ready once the error silencing is removed :)

dhensby:
Oh - and can we get a test, please?

barry-bookinglive:
>>Oh - and can we get a test, please?

Well I'm happy to write one, but I've never been able to run them... it's setup to run them here when I add the pull request right?  else... um...

dhensby:
Yep, they'll run themselves here.

You *might* have to pull your snippet out into another function to test it as I don't think this class is particularly testable at the moment

robbieaverill:
@barry-bookinglive @dhensby any interest in reviving this?

dhensby:
nope :)

PR #37, opened 2015-07-07
Set php command and sort by lastedited

lingo:
This patch was required for a system where a php wrapper is used (for setting up environment variables).
I needed a way to configure this.
Pull request created in case this is useful for others.  If not, then no problem!

### DIFF
--- a/code/tasks/SiteTreeFullBuildEngine.php
+++ b/code/tasks/SiteTreeFullBuildEngine.php
@@ -11,6 +11,12 @@ class SiteTreeFullBuildEngine extends BuildTask {
 	 */
 	protected $urlArrayObject;
 
+	/**
+	 * @var  php_command Command used to run php scripts
+	 * default is 'php'
+	 */
+	protected static $php_command = 'php';
+
 	private static $dependencies = array(
 		'urlArrayObject' =>  '%$URLArrayObject'
 	);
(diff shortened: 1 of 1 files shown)


ss23:
The PHP part looks fine, but can you use shell_escape on the PHP command to prevent people accidently shell injecting themselves (I'd be happy with `set_php_command($command, $escape = true)`.

Why are you including the LastEdited? I don't have objections to it as such, but it has a slight performance impact and I'm not sure why you want it. Can you just explain why? Perhaps a comment in the code saying why you'd do it like that would be useful too.

code/tasks/SiteTreeFullBuildEngine.php
```diff
@@ -137,7 +151,7 @@ protected function getAllLivePages() {
 			Translatable::disable_locale_filter();
 		}		
 		Versioned::reading_stage('Live');
-		$pages = DataObject::get("SiteTree");
+		$pages = DataObject::get("SiteTree")->sort('LastEdited');
```
lingo: This kind of slipped into the commit!  But the rationale was that I wanted to process most recently changed pages first.

dhensby:
Shouldn't this use the config system?

lingo:
I agree that this line shouldn't really be part of the commit, and should use the config system or be otherwise caller-configurable.

lingo:
Do you want me to modify the items mentioned and re-submit a pull request?

dhensby:
Moving to the config system is easy.

Make your `protected static` `private static`; remove your static setter; and check `$this->config()->get('php_command');`

dhensby:
Hmm, is this to get around using something like php56 or php55, etc?

the `sake` util actually gets around this for us out of the box. Maybe we should just be calling sake?

Dan

PR #44, opened 2015-12-04
FIX: Redirect responses written to cache despite empty body.

frankmullenger:
Checking for empty content later in the method, once content for 301/302 redirect responses has been created.

### DIFF
--- a/code/model/FilesystemPublisher.php
+++ b/code/model/FilesystemPublisher.php
@@ -233,17 +233,15 @@ class FilesystemPublisher extends DataExtension {
 			$sanitizedURL = URLArrayObject::sanitize_url($url);
 			$response = Director::test(str_replace('+', ' ', $sanitizedURL));
 
-			// Prevent empty static cache files from being written
-			if (is_object($response) && !$response->getBody()) {
-				SS_Log::log(new Exception('Prevented blank static cache page write for: ' . $path), SS_Log::NOTICE);
-				continue;
-			}
-
-			if (!$response) continue;
-
 			if($response) {
 				$result[$origUrl]['statuscode'] = $response->getStatusCode();
 			}
+			else {
+				// If response object does not exist at all
+				SS_Log::log(new Exception('Response object does not exist for: ' . $path), SS_Log::NOTICE);
+				continue;
+			}
+
 			Requirements::clear();
 			
 			singleton('DataObject')->flushCache();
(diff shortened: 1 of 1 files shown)


dhensby:
Shouldn't we write a php file that will perform the redirect in the cache folder?

frankmullenger:
@dhensby it does fall back to that using OldPageRedirector::onBeforeHTTPError404() but having a statically cached redirect is faster. 

Also, there are quite a few edge cases that OldPageRedirector does not cover. For instance, one query relies on URL segment alone:
https://github.com/silverstripe/silverstripe-cms/blob/3/code/controllers/OldPageRedirector.php#L49

After discussing with @tractorcow, this filter should probably include ParentID = 0 so that it doesn't find other pages nested in the site tree with the same URL segment (an issue we bumped into recently). But there are other edge cases and OldPageRedirector should probably be rewritten. Perhaps we need to denormalize the data in the versions table to look up a page ID for old URLs.

dhensby:
> After discussing with @tractorcow, this filter should probably include ParentID = 0 so that it doesn't find other pages nested in the site tree with the same URL segment (an issue we bumped into recently). But there are other edge cases and OldPageRedirector should probably be rewritten. Perhaps we need to denormalize the data in the versions table to look up a page ID for old URLs.

I've always found this behaviour helpful :/

frankmullenger:
It wouldn't be an alteration of the behaviour to set ParentID = 0 but rather fixing a bug, unless I'm missing something. When the method tries to resolve the URL path into a page ID it recursively calls itself starting with the root node, it doesn't always get that root node if there is more than one page in the site tree with the same URL segment - it gets the first one which might be nested somewhere in the tree (this is the problem we were bumping into).

So if it starts with an incorrect root node it can't find the leaf node and instead of a 301 redirect it continues with the 404. 

Fixing the ParentID for the root node query is a small patch, there are other edge cases that aren't covered. It's actually quite a tricky problem because 1 URL path can potentially resolve to more than 1 page ID over the lifetime of a site.

dhensby:
@frankmullenger at the moment if you have a page on your site which is:

`example.com/page-1/child-page`

but you navigate to

`example.com/child-page`

It will 301 to `example.com/page-1/child-page`

Yes, if there's more than 1 page with the URL segment of `child-page` the first one found will be used; but this isn't a problem as you could never resolve that issue, really.

It may not have been an intended use when this was written, but it's expected given how it's written and that change in behaviour will impact people

In fact, won't adding ParentID = 0 actually mean that the purpose it was intended for won't be matched (ie: this was to redirect sites that had non-nested URLs to nested URLs)? So this behaviour was indeed intended...

PR #54, opened 2016-07-08
Added Option to disable MySQL row-level locking

johannesx75:
Galera Cluster does not support table locking, as they conflict with multi-master replication. If Staticpublishqueue is used in a clustered setup, the lock requests crash the underlying database nodes.

This adds a new configuration option (use_mysql_locks) to control the use of MySQL Locks. By default they are enabled (on MySQL based backends).

To disable row locks add the following to your configuration:
```yml
StaticPagesQueue:
  use_mysql_locks: false
```

### DIFF
--- a/code/model/StaticPagesQueue.php
+++ b/code/model/StaticPagesQueue.php
@@ -56,6 +56,12 @@ class StaticPagesQueue extends DataObject {
 	 *
 	 * @var boolean
 	 */
+	private static $use_mysql_locks = true;
+
+	/**
+	 *
+	 * @var boolean
+	 */
 	private static $realtime = false;
 
 	/**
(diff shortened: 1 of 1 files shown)


edlinklater:
@johannesx75 Has setting this to false been tested? Generally trying to override truthy values with falsey values in the config system doesn't work, see https://docs.silverstripe.org/en/3.3/developer_guides/configuration/configuration/#configuration-values

johannesx75:
Hi @edlinklater ,
yes I just verified again: It works fine on our test system. But that is a 3.1.16 installation. I haven't  tried other versions. As far as I know the overriding is only an Issue if the value has already been defined in another yml file / section.
   
But just to be on the save side, I don't have a problem with turning this around. Would you prefer I we switched the setting to "disable_mysql_locks"? I wasn't sure what the convention for settings is anyway.

dhensby:
@johannesx75 when you've tested did you test by setting it in yml or just in PHP?

If you switch it to `disable_mysql_locks` that'll probably make it a less risky change

johannesx75:
I'll create a new pull request, with disable_mysql_locks.

For completeness, these are the tests I did:
1. Default value in PHP set to true, no setting in yml: works
2. Default value in PHP set to false, no setting in yml: works
1. Default value in PHP set to true, setting in yml changed to false: works
1. Default value in PHP set to true, setting in yml also set to true: works

PR #56, opened 2016-08-25
FIX: PurgeObseleteStaticCacheTask shouldn't clear out index.php

madmatt:
`PurgeObseleteStaticCacheTask` currently has a special case fix for the website homepage to ensure it never deletes that file. However, this doesn't work if the website uses `index.php` to identify the homepage.

I'm not sure if we could use `$fileext` here to be more generic, there's some special magic here around the stale files that I didn't want to mess with.


edlinklater:
Out of the box, staticpublishqueue will never create a cache file with a PHP extension, so this will be custom code for a particular implementation :)

PR #58, opened 2016-12-19
Speed up que object selection

johannesx75:
In a large outstanding que the StaticPagesQueue::get_queue_object can take a long time to select the next record.
This is caused by constantly counting the rows in the StaticPagesQueue. This fix changes that behaviour by asking for the first row instead.


dhensby:
thanks

PR #60, opened 2016-12-22
Fix: setting error fails if record does not exist

johannesx75:
StaticPagesQueue::has_error loads object by url. If the object does not exist (any longer) the code will fail crashing the BuildQue.
Added check to ensure a record was found, otherwise there is no point in setting the error.


dhensby:
thanks

PR #64, opened 2017-03-22
Fixes for PostgreSQL databases

nikrolls:
- PostgreSQL will fail if you try to save a float to an int field
- PostgreSQL needs table and column names to be quoted

### DIFF
--- a/code/reports/BuildStaticCacheSummaryReport.php
+++ b/code/reports/BuildStaticCacheSummaryReport.php
@@ -55,7 +55,7 @@ class BuildStaticCacheSummaryReport extends SS_Report {
 
 			$sort = $field.' '.$direction;
 		} else {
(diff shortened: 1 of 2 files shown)


code/reports/BuildStaticCacheSummaryReport.php
```diff
@@ -55,7 +55,7 @@ function sourceRecords($params, $sort, $limit) {
 
 			$sort = $field.' '.$direction;
 		} else {
-			$sort = $this->dataClass.'.ID DESC';
+			$sort = '"'.$this->dataClass.'"."ID" DESC';
```
dhensby: Can you use `Convert::symbol2sql` to quote columns, please.

nikrolls:
Handy -- thanks!

PR #65, opened 2017-03-22
Fix excludeFromCache not working with SiteTreePublishingEngine

nikrolls:
`URLArrayObject#addObject` adds metadata to the query string of the URL segment that's added to the queue. `#excludeFromCache` passes this query string through to `SiteTree::get_by_link`, which doesn't understand query strings and returns nothing.

There is also an issue with Subsites: SiteTreePublishingEngine disables Subsite filtering, meaning the `SiteTree::get_by_link` will always return the first URL that matches rather than the one indicated by the URL's metadata.

This changes fixes both issues by trying to retrieve the object first using `URLArrayObject#getObject`, and then falling back to a clean (query-string-free) URL passed to `SiteTree::get_by_link`.

### DIFF
--- a/code/model/URLArrayObject.php
+++ b/code/model/URLArrayObject.php
@@ -122,17 +122,18 @@ class URLArrayObject extends ArrayObject {
 	}
 
 	protected function excludeFromCache($url) {
-		$excluded = false;
+		$candidatePage = $this->getObject($url);
 
-		//don't publish objects that are excluded from cache
-		$candidatePage = SiteTree::get_by_link($url);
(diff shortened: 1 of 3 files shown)


dhensby:
Thanks for this and the tests. Would you mind writing out the indentation, please?

nikrolls:
@dhensby, I discovered another issue with this function when it came to Subsites. It's related, so I've updated this PR with an implementation that fixes both issues. I've also fixed the tabs (sorry about that!).

dhensby:
thanks

PR #68, opened 2017-12-01
Try fixing locking issues

jyrkij:
The max-execution-time (30 sec by default) "locking" caused huge
problems on our client sites where rebuild could take far longer than
the max-execution-time specified. This caused new processes to be
spawned although the previous one was still running. Soon there'd be
tens of processes as they all got less and less CPU time... This
solution should take that concern out of the scope.

Tested in two ways:

- crontab entry
- by running one task manually in terminal and trying to run the
  task in another terminal

At least these two ways didn't cause duplicate processes to be spawned.

We are currently running this modified code on two sites and there's at least one other user of the fork: https://stackoverflow.com/a/46316057.

### DIFF
--- a/code/tasks/BuildStaticCacheFromQueue.php
+++ b/code/tasks/BuildStaticCacheFromQueue.php
@@ -2,13 +2,12 @@
 /**
  * This tasks takes care of republishing pages that has been queues in the StaticPagesQueue.
  * 
- * The scripts have a few timers that might need explaining
+ * Only one instace of this script can be run at a time. The script will take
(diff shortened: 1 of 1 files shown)


robbieaverill:
Hi @jyrkij, thank you very much for contributing this pull request, and sorry we haven't reviewed it sooner.

dhensby:
I'm really sorry it's taken so long to look at this. I'm going to merge this so anyone still running ss3 can get a fix for the problem (but given no one else has commented on this, I doubt it's a big problem).

PR #70, opened 2017-12-29
SS4 Compatability

dhensby:
WIP PR for SS4 compatibility.

- [x] Move to queuedjobs module over custom implementation
- [ ] ~~Replace gridfieldajaxreload module with some kind of GridFieldRefreshComponent (if it doesn't exist in core)~~
- [x] Update travis config
- [x] Integrate scrutinizer
- [x] Add CS linting
- [x] Remove references to `.stale` files
- [x] Move deletion to a job
- [x] Handle site tree reorganisations
- [x] Move full static publish to a job
- [x] Use SS templates where needed
- [x] Docs

---

### Nice to haves

- [x] Add codecoverage
- [ ] add Behat tests to test routing
- [ ] Add subsites integration
- [ ] Handle query strings on URLs


edlinklater:
@dhensby I've already done the queuedjobs switch locally.

dhensby:
@edlinklater care to share? :)

andrewandante:
Some context @edlinklater - given that we in the UK office are back from holidays early, we've been asked if we can get this module (and https://github.com/silverstripe/silverstripe-redirectedurls) ready for SS4 by the time the Welly office gets back, as they are part of the upcoming CWP 2.0 package.

andrewandante:
Relies on https://github.com/symbiote/silverstripe-queuedjobs/pull/160

robbieaverill:
I've restarted the builds, looks like different failures now. The queuedjobs exception after phpunit finished can be resolved by disabling the queued jobs shutdown handler function via Config - here's an example: https://github.com/silverstripe/silverstripe-fulltextsearch/blob/415a1b99606affe595c1913d2ba1cfea6450ba39/tests/State/FullTextSearchState.php#L28-L30

andrewandante:
Ta for the QJ fix - not sure why the other test is failing, passes locally. Will push some awkward var_dumps to Travis 😬

andrewandante:
Refactored the tests to not use `Director::test()` as that invokes SS and basically ignores the whole point of static publishing. We'll need some behat tests of some form to check the routing, but we can at least test that the files are successfully created.

dhensby:
OK - we've made a fairly big refactor of the caching strategy and how one would implement PHP based caching.

In the SS3 version error responses are cached in `cache/error-[code].html` - this doesn't provide much benefit for pages that need to be cached and have an error response code (eg: maybe there's a URL that gets hit a lot (favicon.ico or autodiscover.xml) and you want to cache that 404 response. At the moment those wouldn't be intercepted by the current .htaccess rules of `URL -> cache/url.html` and they also wouldn't respond with the correct response code even if they did (they'd give you 200 response codes, even under PHP caching because the response code isn't correctly cached in the PHP caches).

Therefore we've decided to move to store response headers and codes in the php cache file as well as the error response in the matching .html file. (only the html file is stored if you've opted for HTML only caching).

If the response to be cached is an error response then we'll only cache it if we're using PHP based caching because this is the only way we can guarantee the correct response code - we'll just mess up SEO and whatnot otherwise by having apache serve 200 responses for error pages.

This approach also allows us to simplify how we serve the static PHP caches - rather than complex routing in the `.htaccess` we can now add some simple code to the project `index.php` instead:

```diff
<?php
use SilverStripe\Control\HTTPApplication;
use SilverStripe\Control\HTTPRequestBuilder;
use SilverStripe\Core\CoreKernel;
use SilverStripe\Core\Startup\ErrorControlChainMiddleware;

+$requestHandler = require 'vendor/silverstripe/staticpublishqueue/includes/staticrequesthandler.php';
+
+// successful cache hit
+if (false !== $requestHandler('cache')) {
+    die;
+} else {
+    header('X-Cache-Miss: ' . date(\Datetime::COOKIE));
+}
+
require __DIR__ . '/vendor/autoload.php';

// Build request and detect flush
$request = HTTPRequestBuilder::createFromEnvironment();

// Default application
$kernel = new CoreKernel(BASE_PATH);
$app = new HTTPApplication($kernel);
$app->addMiddleware(new ErrorControlChainMiddleware($app));
try {
    $response = $app->handle($request);
    $response->output();
} catch (Exception $e) {
    $e->getTraceAsString();
    throw $e;
}
```

Caching both a PHP and HTML file (PHP file stores the header and responsecode meta data, HTML file stores the response body), we can also allow server admins to quickly switch to HTML based caching when their servers are under high load without having to rebuild the static cache.

Overall we've taken a measured approach between raw speed (just using apache to serve HTML files) and SEO compliance (having correct response codes). The way this works now is that people will see a significant performance increase from a non-statically published site but they won't see the absolute best performance that they would from HTML (but they also avoid the drawbacks/pitfalls.

edlinklater:
@dhensby Looks good! We'll need to document what features you do & don't get in each of the strategies (PHP vs HTML) and it would be good to still provide an example `.htaccess` / nginx rewrite for those who want to run an entirely HTML setup.

dhensby:
> @dhensby Looks good! We'll need to document what features you do & don't get in each of the strategies (PHP vs HTML) and it would be good to still provide an example .htaccess / nginx rewrite for those who want to run an entirely HTML setup.

for sure - we'll get the docs written as one of the last steps - at the moment it's all in the README.md and really I think some better structured docs would help...

src/Job/StaticCacheFullBuildJob.php
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * @return array
+     */
+    protected function getAllLivePageURLs()
+    {
+        $urls = [];
+        $livePages = Versioned::get_by_stage(SiteTree::class, Versioned::LIVE);
+        foreach ($livePages as $page) {
+            if ($page->hasExtension(PublishableSiteTree::class) || $page instanceof StaticallyPublishable) {
+                $urls = array_merge($urls, $page->urlsToCache());
```
dhensby: Not sure that `array_merge` will nicely merge keyed arrays - check this.

src/Job/StaticCacheFullBuildJob.php
```diff
@@ -0,0 +1,99 @@
...
+            }
+        }
+        // @TODO look here when integrating subsites
+        // if (class_exists(Subsite::class)) {
+        //     Subsite::disable_subsite_filter(true);
+        // }
+
+        return array_unique($urls);
```
dhensby: array_unique is not needed / will break things. this array is `$url => $priority`

src/Publisher/FilesystemPublisher.php
```diff
@@ -206,4 +206,48 @@ protected function URLtoPath($url)
...
+
+    protected function pathToURL($path)
+    {
+        if (strpos($path, $this->getDestPath()) === 0) {
+            //Strip off the full path of the cache dir from the front
+            $path = substr($path, strlen($this->getDestPath()));
+        }
+        $path = ltrim($path, '/');
+
+        if (strpos($path, $this->getDestFolder()) === 0) {
```
dhensby: elseif?

src/Publisher/FilesystemPublisher.php
```diff
@@ -206,4 +206,48 @@ protected function URLtoPath($url)
...
+    public function getPublishedURLs($dir = null, &$result = [])
+    {
+        if ($dir == null) {
+            $dir = $this->getDestPath();
+        }
+
+        $root = scandir($dir);
+        foreach($root as $fileOrDir)
+        {
+            if ($fileOrDir === '.' || $fileOrDir === '..') {
```
dhensby: if first char is `.`

src/Publisher/FilesystemPublisher.php
```diff
@@ -206,4 +206,48 @@ protected function URLtoPath($url)
...
+                continue;
+            }
+            $fullPath = $dir . DIRECTORY_SEPARATOR . $fileOrDir;
+            if (is_file($fullPath)) {
```
dhensby: if extension is `html` and is_file?

src/Task/StaticCacheFullBuildTask.php
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * Implement this method in the task subclass to
+     * execute via the TaskRunner
+     *
+     * @param HTTPRequest $request
+     * @return
+     */
+    public function run($request)
+    {
+        $job = new StaticCacheFullBuildJob();
```
dhensby: would be good to queue this job

andrewandante:
Thinking about subsite integration - maybe the best way would be to have an intermediary module? That way we could use Injector to have a `SubsitesFileSystemPublisher` or whatever. It feels gross to tightly couple to Subsites in this module by itself.

src/Job/StaticCacheFullBuildJob.php
andrewandante: > If the input arrays have the same string keys, then the later value for that key will overwrite the previous one. If, however, the arrays contain numeric keys, the later value will not overwrite the original value, but will be appended.

src/Task/StaticCacheFullBuildTask.php
andrewandante: 🥇 😤

src/Publisher/FilesystemPublisher.php
```diff
@@ -234,11 +228,12 @@ public function getPublishedURLs($dir = null, &$result = [])
...
         $root = scandir($dir);
         foreach ($root as $fileOrDir) {
-            if ($fileOrDir === '.' || $fileOrDir === '..') {
+            if (strpos($fileOrDir, '.') === 0) {
                 continue;
             }
             $fullPath = $dir . DIRECTORY_SEPARATOR . $fileOrDir;
-            if (is_file($fullPath)) {
+            // we know html will always be generated, this prevents double ups
+            if (is_file($fullPath) && pathinfo($fullPath)['extension'] == 'html') {
```
dhensby: `pathinfo($fullPath, PATHINFO_EXTENSION)` feels a little neater here

src/Task/StaticCacheFullBuildTask.php
```diff
@@ -2,22 +2,96 @@
...
+     * Queue up a StaticCacheFullBuildJob
+     * Check for startAfter param and do some sanity checking
      *
      * @param HTTPRequest $request
      * @return
      */
     public function run($request)
     {
         $job = new StaticCacheFullBuildJob();
+        $signature = $job->getSignature();
```
dhensby: The signature can be effected by the job's `jobData`, or not for the full build job?

dhensby: I'd consider doing something like:

```php
$ids = JobDescriptor::get()->getIDList();
$id = QueuedJobService::singleton()->queueJob($job);
if (in_array($id, $ids)) {
    return "already in queueu";
}
```

Rather than duplicating some of the `QueuedJobService` duplicate detection logic

edlinklater:
@andrewandante Once you're touching Subsites you need to consider things like knowing which domains point to which Subsite. In order to avoid spinning up SS and/or doing DB requests each time, you're probably looking at building out a map file that PHP can parse very quickly each lookup.

andrewandante:
@edlinklater I am 100% certain that Subsites is a giant minefield - part of the reason I want to leave it out for now and have a bridging module that can deal with it. Same with reports, although I suspect they'll be less complicated.

Otherwise, we are getting pretty close to a working implementation - just writing tests at the moment for things like changing a URLSegment and SiteTree re-organisation, and some behat-style tests to ensure that we are hitting cached versions of URLs. And docs 😄

dhensby:
new dependency: https://github.com/silverstripe/silverstripe-framework/pull/7746

andrewandante:
New new dependency: https://github.com/silverstripe/silverstripe-versioned/pull/82

andrewandante:
These new dependencies have broken the tests - but they are all passing locally 😁

andrewandante:
OK well these tests are passing, but Scrutinzer is having a meltdown over `composer-plugin-api` for some reason and Travis won't pull in the correct version of Versioned 😤

robbieaverill:
You can switch the scrutinizer config to their new build engine to fix the composer problem- see other vendor modules for an example, e.g. taxonomy

andrewandante:
@robbieaverill ta - although taxonomy hasn't been updated 😁 I used subsites as my pasta to copy

andrewandante:
@edlinklater what email would you like as the maintainer email? I'm updating the docs and procrastinating 😝

guci0:
It's good to go?

andrewandante:
@guci0 I would consider it in alpha :) absolutely go ahead and pull it in and try it out, we'd love to know what doesn't work :)

guci0:
Something for cleaning ("exclude"?) meta names is needed, I have
<meta name="x-page-id" content="3" />
<meta name="x-cms-edit-link" content="http://localhost/2017/test.com/admin/pages/edit/show/3" />
// isn't it from CMS view, split view?
and for url handling?  for specific port ex. localhost:8888 :) for .html (cache). Let them work with assets. Base href?

dhensby:
@guci0 Are you building the pages through the queue? I'm not seeing the CMS meta tags coming through on my statically published pages.

Getting the right port in the `<base>` tag isn't working and (I believe) doesn't work in the current stable version either.

andrewandante:
@guci0 once [this PR](https://github.com/silverstripe/silverstripe-framework/pull/7780) is merged you should be able to set `alternate_base_url` to fix the base_tag problems - until then it won't work, and hasn't previously.

andrewandante:
@robbieaverill what's the next step to get this merged? It'd be great to be able to tag an alpha at least. I know @frankmullenger is having a bash around with this branch at the moment, but it would be good to have something for the wider public to find holes in 😄

dhensby:
Yep - I'm happy with the current feature set and it'd be good to start letting people battle harden this a little :)

robbieaverill:
@dhensby since you’ve been more involved that I have in this one perhaps you could review/merge?

dhensby:
@robbieaverill I wrote a lot of it!

I can (will) re-review, but it'll still be a self merge/review

dhensby:
Oh, and FYI - this is more a re-write so it's best to just review the code as is than via a diff, which is just going to be serious noise

chrispenny:
@dhensby just FYI I'm implementing this into the TNZ project atm.

We have Assets attached to Blocks (which are attached to Pages), so I've added a new `Trigger` and `Engine` onto `File`, and so far that all seems to be working perfectly.

I haven't quite had time to conclusively test all the Fluent considerations, but I don't expect to find any issues (since it just uses a URL to generate the cache, and the URL has the Locale in it).

I'm hoping to be done EOD tomorrow, so I'll let you know how it goes.

Thank you for all your effort so far, it's looking really good :)

dhensby:
@chrispenny How's it gone? If the locale is a query string then you may have some problems out of the box as query strings are not supported (yet)

dhensby:
Is there a year we can add for the latest commit?

andrewandante:
Will look at it - was responding to feedback from @guci0 who demonstrated that no theme was being picked up when he ran the job manually from the QueuedJobs Admin

chrispenny:
Oops! Sorry, @dhensby. I took the full Waitangi weekend off - forgot to update beforehand.

For the most part, our implementation has been smooth. I didn't notice any issues using this with Fluent - it all just worked.

I did personally somehow get into a weird cached state where my browser had (potentially) cached the static page for a `?stage=Stage` response, but I couldn't replicate it with any consistency, so I'll raise that later down the line if I manage it again.

Awesome work :)

blueo:
@dhensby just an update on @chrispenny's comment above. If you have a fresh session and log in without first visiting an page that calls `Versioned::choose_site_stage` you end up without a 'bypassStaticCache` cookie. This means that when you do try view the draft of a cached page, you can never see the draft content. The cached file will be returned until either you visit a page that calls 'choose_site_stage' or delete the cache  (not an option on deployed sites unfortunately) allowing 'choose_site_stage' to run on the next visit. I can reproduce it by caching a page, then clearing cookies and logging back in via `/admin`, now I can't see the draft of the page.

chrispenny:
@dhensby, I managed to replicate with the following steps. I used Chrome for publishing, and Safari for browsing. I haven't used Safari on this machine before, so it is completely "clean".

Chrome:
1) Create new Page and publish.
2) Wait for static file to be generated.
3) Make a change to the Page and save as draft.

Safari:
1) Log into CMS (do not browse to any "page previews").
2) Browse to Page on the frontend **without** `?stage=Stage`.
3) Browse to Page on the frontend **with** `?stage=Stage`.
4) Browse to Page in CMS edit form and enter "preview" or "split" mode.

Result:
I can't see the uncached version of this file (`?stage=Stage`, or `?stage=Live`) - either on the frontend, or from within the CMS preview window.

Our content authors have been reporting that "Pages aren't updating", and I suspect they're experiencing a similar issue to the above.

CC: @frankmullenger

dhensby:
@chrispenny @frankmullenger @blueo 

This sounds pretty odd and I think @chrispenny is describing as the steps to replicate are different to what the content editors are experiencing.

It sounds like the content editors are simply not viewing the stage content, so of course they won't see it. If they've never gone to a page with `?stage=Stage` in the URL they'll never see the stage content, even if they do bypass the static cache.

However, I have noted (and pushed a fix for) a problem where the query strings were not being ignored in a way that meant static caches were bypassed (as they used to be).

dhensby:
My fix will have resolved the issue that @chrispenny has raised which is that appending `?stage=Stage` to a statically published paged never invoked the change of stage.

However, if you never ever go to a URL with `?stage=Stage` in it, you'll never see stage content.

chrispenny:
@dhensby Thanks for that. I'll pull the change down today and see how it goes.

To clarify - I didn't mean that the content authors **never** append `?stage=Stage`, just that it might not be present the first time they viewed the page.

@frankmullenger can you please chime in on whether bypassing cache when query strings are present is going to be an issue? A/B testing I suspect will be fine, but potentially targeting will be an issue(?). Maybe we can use hash values though - I'm not sure.

frankmullenger:
As @dhensby points out this change brings the default behaviour in line with the previous version so that seems reasonable.  

As a requirement, we (and other projects) sometimes need a way to serve the static cache when GET params are passed. So there should be a way to change this behaviour and it appears that we can do that by passing a function as the argument for $urlMapping or by altering the .htaccess rules for serving static files depending on how the module is configured.

dhensby:
I feel that query string support *should* be supported OOTB for this module, but it's beyond the scope of upgrading the module for SS4 compat.

You can override the `URLtoPath` method to take into account query strings if you need it in your own project. :)

chrispenny:
Sounds good to me.

We haven't come across any issues since the update, so I feel like everything is good to go!

andrewandante:
Current test failures are linting failures...

_glares at Dan_

dhensby:
The failures are so bizarre, I've got it locally and there are not 4 lines after the use statements...

I rant the `phpcbf` bin over the code and it did nothing...

andrewandante:
Yeah that is super weird. I'm likely going to push up another commit today, so hopefully that will re-trigger and wipe it?

src/Publisher/FilesystemPublisher.php
```diff
@@ -191,10 +192,16 @@ protected function pathToURL($path)
...
         }
         $path = ltrim($path, '/');
+
         // Strip off the file extension
         $relativeURL = substr($path, 0, (strrpos($path, ".")));
 
-        return $relativeURL == 'index' ? '/' : $relativeURL;
+        if (FilesystemPublisher::config()->get('domain_based_caching')) {
+            // factor in the domain as the top dir
+            return Director::protocol() . $relativeURL;
```
dhensby: I'm confused here - why are we concatenating the protocol to the relative url?

robbieaverill:
>The failures are so bizarre, I've got it locally and there are not 4 lines after the use statements... I rant the phpcbf bin over the code and it did nothing...

Just got this on another project when pulling dev-master for phpcs. Make sure you've got a stable constraint and `prefer-stable: true` in composer

andrewandante:
@robbieaverill yeah we figured it out: https://github.com/squizlabs/PHP_CodeSniffer/pull/1899

dhensby:
@robbieaverill we fixed PHPCS! whoop

dhensby:
OK - I've gone over this once again and made a few final changes.

I'm happy as this is :)

src/Job/StaticCacheFullBuildJob.php
```diff
@@ -0,0 +1,100 @@
...
+ * Adds all pages to the queue for caching. Best implemented on a cron via StaticCacheFullBuildTask.
+ */
+class StaticCacheFullBuildJob extends Job
+{
+    /**
+     * @return string
+     */
+    public function getTitle()
+    {
+        return 'Generate a static pages for all URLs';
```
robbieaverill: This title is pretty similar to GenerateStaticCacheJob - maybe could use clarification

src/Job/StaticCacheFullBuildJob.php
```diff
@@ -0,0 +1,100 @@
...
+            if ($page->hasExtension(PublishableSiteTree::class) || $page instanceof StaticallyPublishable) {
+                $urls = array_merge($urls, $page->urlsToCache());
+            }
+        }
+
+        $this->extend('afterGetAllLivePageURLs', $urls);
+        // @TODO look here when integrating subsites
+        // if (class_exists(Subsite::class)) {
+        //     Subsite::disable_subsite_filter(true);
+        // }
```
robbieaverill: Do you want to do this now or later?

src/Publisher.php
```diff
@@ -0,0 +1,151 @@
...
+                    'REMOTE_ADDR' => '127.0.0.1',
+                    'HTTPS' => $urlParts['scheme'] == 'https' ? 'on' : 'off',
+                    'QUERY_STRING' => isset($urlParts['query']) ? $urlParts['query'] : '',
+                    'REQUEST_TIME' => DBDatetime::now()->getTimestamp(),
+                    'REQUEST_TIME_FLOAT' => (float) DBDatetime::now()->getTimestamp(),
+                    'HTTP_HOST' => $urlParts['host'],
+                    'HTTP_USER_AGENT' => 'silverstripe/staticpublishqueue',
+                ],
+                '_GET' => $getVars,
+                '_POST' => [],
```
robbieaverill: Indentation is a bit out

src/Publisher.php
```diff
@@ -0,0 +1,151 @@
...
+        }
+        return $response;
+    }
+
+    /**
+     * @return HTTPApplication
+     */
+    protected function getHTTPApplication()
+    {
+        $kernel = new CoreKernel(BASE_PATH);
```
robbieaverill: Probably better to use `Injector::inst()->create(Kernel::class, BASE_PATH)` unless there's a reason to skip the injector

src/Publisher/FilesystemPublisher.php
```diff
@@ -0,0 +1,232 @@
...
+            'url' => $url,
+            'path' => false,
+        ];
+    }
+
+    /**
+     * @param string $url
+     * @return array A result array
+     */
+    public function publishURL($url, $forcePublish = false)
```
robbieaverill: Up for debate, but I'd rather see a PSR-2 compatible `publishUrl()` method name than a semantically correct `publishURL()` method name

tests/php/FilesystemPublisherTest.php
```diff
@@ -0,0 +1,360 @@
...
+            'parent/child',
+            $urlToPath->invokeArgs($this->fsp, ['parent/child'])
+        );
+    }
+
+    public function testUrlToPathWithAbsoluteUrls()
+    {
+        $reflection = new \ReflectionClass(FilesystemPublisher::class);
+        $urlToPath = $reflection->getMethod('URLtoPath');
+        $urlToPath->setAccessible(true);
```
robbieaverill: This opens up the debate about whether we should test protected methods: https://stackoverflow.com/questions/249664/best-practices-to-test-protected-methods-with-phpunit

There's a few cases of this in this integration test class - perhaps we should just make the method public and be done with it?

src/Publisher.php
dhensby: Kernels can't be injected because they are created before the Manifests / Injector is ready (the Kernels create the injector instances) - so whilst we could use the injector, it wouldn't make too much sense, it'd be better for someone to create their own Publisher and override the `getHTTPApplication` method IMO

src/Publisher/FilesystemPublisher.php
dhensby: Our coding convention has been to capitalise abbreviations for quite some time.

PSR-2 gives no opinion on this from what I can read: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md

tests/php/FilesystemPublisherTest.php
dhensby: We only want to expose public methods for `StaticPublisher` as that's the interface we are implementing. `URLtoPath` is very much an internal method for the `FilesystemPublisher` which doesn't make sense in the context of an abstract "publisher" engine.

Making it public will mean people make assumptions about it being available on all Publishers and that's not accurate.

We want to test the internal logic of this class whilst not exposing the methods for all to see. Whilst I can relate to the argument that we should test the methods that rely on this method, that provides implicit coverage whereas we want to provide some explicit coverage of this method.

The alternative is just to remove the tests IMO

src/Job/StaticCacheFullBuildJob.php
dhensby: Later - subsites is on our "nice to have list" and we need to get this merged and move on to other work

dhensby:
@robbieaverill I've implemented some of the feedback and left responses on others :)

src/Job/StaticCacheFullBuildJob.php
andrewandante: Will be a connector module eventually :)

dhensby:
MERGED!

Issue #78, opened 2018-05-16
Feature request:  add timestamp to published html


DorsetDigital:
Whilst adding the static publisher to a site, and particularly whilst developing additional code around it, I found it very useful to be able to see when pages were actually being updated (or not as the case may be).

To get around this, I put a simple hack in place which adds the current timestamp to the end of the HTML as a comment:

In `FilesystemPublisher::publishPage()`:

```
protected function publishPage($response, $url)
    {
        $success = true;
        if ($path = $this->URLtoPath($url)) {
            if ($this->getFileExtension() === 'php') {
                $phpContent = $this->generatePHPCacheFile($response);
                $success = $this->saveToPath($phpContent, $path . '.php');
            }
            $responseBody = $response->getBody();
            $responseBody = str_replace('</html>', "<!-- ".date('d/m/Y H:i:s')." -->\n</html>", $responseBody);            
            return $this->saveToPath($responseBody, $path . '.html') && $success;
        }
        return false;
    }
```

A few possible questions:

- Should this be active in dev mode (and dev mode only)?
- Should this be a config option to enable?
- Is there a better way to add the timestamp?

dhensby:
I think pushing this into the cached output isn't a great idea. We don't actually have any knowledge that the output is indeed HTML and also adding comments outside the `</html>` is invalid markup, I believe.

I'd go for an approach of adding a header to the response instead. `X-Cached-At` or something... You could extend the default `Publisher` and add the header in `generatePHPCacheFile`.

Are you thinking of creating a PR for this?

DorsetDigital:
I'd be happy to do a PR if we can reach a consensus on the best method.
The issue I can see with adding it to the PHP file is that it might not be generated - this was the case for me.   I had only HTML files enabled, so this was the simplest way.

I take your point about the output possibly not being HTML, in the case of JSON, XML, etc. it wouldn't work - that's why I used a `str_replace()` on the closing HTML tag...  if it wasn't there, it should fail gracefully.   The comments appear before the closing tag, so should still be valid markup

crlindsay2:
instead of injecting it into the html, you could write a LastUpdated (or whatever you want to call it) value before publishing, then have an if statement in the <head> of your doc with the  inside. That way it *should* only be inserted and visible if it's static published, and you're not injecting html.

In theory anyway, i don't know offhand if you can change a db value and have it reflected in the template as part of the publish function.

DorsetDigital:
[...]

PR #79, opened 2018-05-16
Add optional timestamps to cached HTML

DorsetDigital:
As per #78 
Add code to FilesystemPublisher to deal add timestamps

### DIFF
--- a/src/Publisher.php
+++ b/src/Publisher.php
@@ -42,6 +42,13 @@ abstract class Publisher implements StaticPublisher
     private static $domain_based_caching = false;
 
     /**
+     * @config
+     *
+     * @var Boolean Add a timestamp to the statically published output for HTML files
+     */
+    private static $add_timestamp = false;
+
+    /**
      * @param string $url
      *
      * @return HTTPResponse
@@ -92,6 +99,10 @@ abstract class Publisher implements StaticPublisher
             );
             $app = $this->getHTTPApplication();
             $response = $app->handle($request);
+
+            if ($this->config()->get('add_timestamp') === true) {
+                $response->setBody(str_replace('</html>', "<!-- " . DBDateTime::now() . " -->\n</html>", $response->getBody()));
+            }
         } catch (HTTPResponse_Exception $e) {
             $response = $e->getResponse();
         } finally {
--- a/src/Publisher.php
+++ b/src/Publisher.php
@@ -100,11 +100,11 @@ abstract class Publisher implements StaticPublisher
             $app = $this->getHTTPApplication();
             $response = $app->handle($request);
 
-            if ($this->config()->get('add_timestamp') === true) {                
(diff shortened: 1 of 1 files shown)


src/Publisher/FilesystemPublisher.php
```diff
@@ -148,7 +167,12 @@ protected function publishPage($response, $url)
                 $phpContent = $this->generatePHPCacheFile($response);
                 $success = $this->saveToPath($phpContent, $path . '.php');
             }
-            return $this->saveToPath($response->getBody(), $path . '.html') && $success;
+            $responseBody = $response->getBody();
+            if ($this->getAddTimestamp() === true) {
+                $responseBody = str_replace('</html>', "<!-- " . date('c') . " -->\n</html>", $responseBody);
```
dhensby: Can we use `DateTime::now()` please as this will be testable

dhensby:
Thanks for this, how comes you're not using a config setting?

Please can you also add some docs for this (and squash your current commits)?

DorsetDigital:
I figured it was more 'Stripey' to be able to set it in the yml file (similar to the file extension option).  Is that not the best way?  Happy to take guidance on this.

dhensby:
> I figured it was more 'Stripey' to be able to set it in the yml file (similar to the file extension option). Is that not the best way? Happy to take guidance on this.

oh, setting it through injector params? I guess we could do it that way.

Another point though, can you make this change in `Publisher@generatePageResponse` and update the body of the response so that any/all publishers have this added rather than just the filesystempublisher?

DorsetDigital:
I've moved the functionality into the Publisher class and swapped to a config-based approach, it seemed like a simpler way to do it with hindsight.
I'll add something to the docs now

DorsetDigital:
I've now squashed all the commits as well, so hopefully I didn't break anything in the process!

src/Publisher.php
```diff
@@ -92,6 +99,10 @@ public function generatePageResponse($url)
             );
             $app = $this->getHTTPApplication();
             $response = $app->handle($request);
+
+            if ($this->config()->get('add_timestamp') === true) {
+                $response->setBody(str_replace('</html>', "<!-- " . DBDateTime::now() . " -->\n</html>", $response->getBody()));
```
dhensby: `DBDateTime::now()` needs to be formatted appropriately eg: `DBDateTime::now()->Full()` or maybe `DBDateTime::now()->Rfc2822()`

dhensby: You also need to bring this line below 120 chars long

DorsetDigital:
Are there still changes needed for this to be merged?   (I just wanted to make sure you're not waiting on me to do anything!)

dhensby:
sorry, fell off my radar - looks acceptable to me

PR #91, opened 2018-10-05
FIX FullBuildJob processing same URLs multiple times

scott1702:
I ran into an issue performing a Full build of a site with a lot of URLs, as the job hits the chunk_size and re-initialises it seems like it doesn't retain the `URLsToProcess` of the previous run of the job. The job then starts building the cache as if it hadn't processed any URLs, so hits the chunk_size again and just gets stuck in a loop of rebuilding the first few URLs but never finishing the job.


robbieaverill:
Makes sense

robbieaverill:
Postgres failure is unrelated to this change, and can be resolved by bumping to 2.1.x-dev or higher in travis.yml for that module

PR #93, opened 2018-10-08
FIX FilesystemPublisher now writes to a temporary file then moves it to the desired path

### LINKED ISSUE #90: Lock files before writing, avoid empty responses
`FilesystemPublisher->saveToPath()` uses `file_put_contents()`. For a very short time, this will produce an empty file while the blocks are being written. If the file is read by another process (PHP, Apache, nginx, etc), it might return the empty or partially written file. See https://tqdev.com/2018-locking-file-cache-php.

We should write to a temporary file before renaming to the original it to ensure atomicity under high concurrency.

/cc @frankmullenger 

## Pull requests 

- [ ] #93

robbieaverill:
This prevents issues where file_put_contents() does not request a file lock during writes, so processes may pick up an empty file during the process. Instead, we write to a system temp directory and rename it at the end to the desired filename.

Also renames the test class for the publisher so it makes a predictable namespace.

Fixes #90

### DIFF
--- a/src/Publisher/FilesystemPublisher.php
+++ b/src/Publisher/FilesystemPublisher.php
@@ -163,9 +163,18 @@ class FilesystemPublisher extends Publisher
         if (empty($content)) {
             return false;
         }
+
+        // Write to a temporary file first
+        $temporaryPath = tempnam(TEMP_PATH, 'filesystempublisher_');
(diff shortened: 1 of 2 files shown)


micmania1:
Looks good to me. I don't have merge rights though :)

PR #96, opened 2018-11-07
Ensure URLs get queued when BASE_URL is set

stojg:
when BASE_URL is set, which is pretty much a must when running on https, URLs won't get queued due to something that I presume is a very old and legacy check.


robbieaverill:
PHP 5.3 is gone in Travis, and this has been switched to 5.6 in the most recent commit to the 3 branch

PR #105, opened 2019-07-30
Bugfix/domain based caching

micmania1:
This includes a few things:
1. Fixes a bug where the index files are deleted during a full cache rebuild when using domain based caching
2. Added a simple property to force SSL caching. At the moment there's cases where you can build a cache with HTTP and when accessing over HTTPS, nothing loads (font, css, images etc.) due to insecure content. This can be the case when running on the CLI.
3. Allow dev to replace StaticCacheFullBuildJobs with their own implementation
4. Fixes a bug where currentStep is not incremented meaning that when the no. of URLs to process are greater than `chunk_size`, the job will reinitialise and start again, failing to move forward. The job is stuck in a 'waiting' state.

### DIFF
--- a/docs/en/basic_configuration.md
+++ b/docs/en/basic_configuration.md
@@ -1,7 +1,7 @@
 # Basic Configuration
 
-By default, the extensions `SiteTreePublishingEngine` and `PublishableSiteTree` 
-are applied to `SiteTree`, so as soon as the module is installed, it should 
+By default, the extensions `SiteTreePublishingEngine` and `PublishableSiteTree`
+are applied to `SiteTree`, so as soon as the module is installed, it should
 be ready to go.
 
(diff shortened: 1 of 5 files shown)


emteknetnz:
@micmania1 looks like this PR may be merged with a little work, though first we'd need this rebased/retargeted to `3.3` instead of `master` so that this can be released as a patch.  This should also resolve the merge conflicts and hopefully also resolve travis being red.  Cheers.

dhensby:
If @micmania1 isn't able to, I don't mind having a look at rebasing this to see how much work it is to get it green

micmania1:
@dhensby go for it :+1:  I won't get time to look into this any time soon

dhensby:
Rebased and green :) 💚

PR #108, opened 2020-02-12
Expose path to URL

mfendeksilverstripe:
# Expose PathToURL function

It's quite handy to have the `PathToURL` available especially when building a better UI for static cache manipulation. We already have `URLtoPath` available, so this change-set simply exposes the `PathToURL` function as well. There should be no functional changes.

## Related issues
https://github.com/silverstripe/silverstripe-staticpublishqueue/issues/109


chillu:
Looks good to me, thanks Mojmir!

PR #115, opened 2020-05-21
BUGFIX: PHP `Notice: Undefined index: extension` if there's no file extension

mattclegg:
Updating to use PATHINFO_EXTENSION as this doesn't `Notice` if there's no extension.

Ref; https://www.php.net/manual/en/function.pathinfo.php#example-2967


mattclegg:
@emteknetnz Any idea how to restart the Scrutinizer test? It looks like it failed to checkout.

emteknetnz:
Red build is scrutinizer having a fairly random failure, I cannot restart build.  Travis is green.

PR #118, opened 2020-07-05
adding gzip option

sunnysideup:
It is up to maintainers if you are interested in this, but I can report that it works for me with about 20% speed gains and lower server profile.


composer.json
```diff
@@ -1,12 +1,15 @@
 {
-    "name": "silverstripe/staticpublishqueue",
+    "name": "sunnysideup/staticpublishqueue",
```
dhensby, line 2: We can't change this

docs/examples/.htaccess.example-for-gzip
```diff
@@ -0,0 +1,129 @@
...
+
+
+
+<IfModule mod_rewrite.c>
+	SetEnv HTTP_MOD_REWRITE On
+	RewriteEngine On
+
+	## CONFIG FOR STATIC PUBLISHING
+	# Cached content - sub-pages (site in the root of a domain)
+    RewriteCond "%{HTTP:Accept-encoding}" "gzip"
```
dhensby, line 37:
```suggestion
	RewriteCond "%{HTTP:Accept-encoding}" "gzip"
```

docs/examples/.htaccess.example-for-gzip
```diff
@@ -0,0 +1,129 @@
...
+	RewriteCond %{REQUEST_METHOD} ^GET|HEAD$
+	RewriteCond %{QUERY_STRING} ^$
+	RewriteCond %{REQUEST_URI} /(.*[^/])/?$
+	RewriteCond %{DOCUMENT_ROOT}/cache/%1.html.gz -f
+	RewriteCond %{REQUEST_FILENAME} !-f
+	RewriteRule .* /cache/%1.html.gz [L]
+
+	# Cached content - homepage (site in root of a domain)
+    RewriteCond "%{HTTP:Accept-encoding}" "gzip"
```
dhensby, line 46:
```suggestion
	RewriteCond "%{HTTP:Accept-encoding}" "gzip"
```

docs/examples/.htaccess.example-for-gzip
```diff
@@ -0,0 +1,129 @@
...
+	RewriteCond %{QUERY_STRING} ^$
+	RewriteCond %{REQUEST_URI} ^/?$
+	RewriteCond %{DOCUMENT_ROOT}/cache/index.html.gz -f
+	RewriteCond %{REQUEST_FILENAME} !-f
+	RewriteRule .* /cache/index.html.gz [L]
+
+	# Config for site in a subdirectory
+	# (START remove the following four rules if site is on root of a domain. E.g. test.com rather than test.com/my-site)
+	# Cached content - sub-pages (site in sub-directory)
+    RewriteCond "%{HTTP:Accept-encoding}" "gzip"
```
dhensby, line 57:
```suggestion
	RewriteCond "%{HTTP:Accept-encoding}" "gzip"
```

docs/examples/.htaccess.example-for-gzip
```diff
@@ -0,0 +1,129 @@
...
+	RewriteCond %{REQUEST_METHOD} ^GET|HEAD$
+	RewriteCond %{QUERY_STRING} ^$
+	RewriteCond %{REQUEST_URI} /(.*[^/])/(.*[^/])/?$
+	RewriteCond %{DOCUMENT_ROOT}/%1/cache/%2.html.gz -f
+	RewriteCond %{REQUEST_FILENAME} !-f
+	RewriteRule .* cache/%2.html.gz [L]
+
+	# Cached content - homepage (site in sub-directory)
+    RewriteCond "%{HTTP:Accept-encoding}" "gzip"
```
dhensby, line 66:
```suggestion
	RewriteCond "%{HTTP:Accept-encoding}" "gzip"
```

includes/functions.php
```diff
@@ -2,22 +2,24 @@
 
 namespace SilverStripe\StaticPublishQueue;
 
-if (!function_exists('SilverStripe\\StaticPublishQueue\\URLtoPath')) {
+if (! function_exists('SilverStripe\\StaticPublishQueue\\URLtoPath')) {
```
dhensby, line 5:
```suggestion
if (!function_exists('SilverStripe\\StaticPublishQueue\\URLtoPath')) {
```

includes/functions.php
```diff
@@ -2,22 +2,24 @@
...
+        $url = str_replace('?stage=Live', '?', $url);
+
         // parse_url() is not multibyte safe, see https://bugs.php.net/bug.php?id=52923.
         // We assume that the URL hsa been correctly encoded either on storage (for SiteTree->URLSegment),
         // or through URL collection (for controller method names etc.).
         $urlParts = @parse_url($url);
 
         // query strings are not yet supported so we need to bail is there is one present
-        if (!empty($urlParts['query'])) {
+        if (! empty($urlParts['query'])) {
```
dhensby, line 16:
```suggestion
        if (!empty($urlParts['query'])) {
```

includes/functions.php
```diff
@@ -26,10 +28,10 @@ function URLtoPath($url, $baseURL = '', $domainBasedCaching = false)
         // Normalize URLs
         $urlSegment = trim($urlSegment, '/');
 
-        $filename = $urlSegment ?: "index";
+        $filename = $urlSegment ?: 'index';
 
         if ($domainBasedCaching) {
-            if (!$urlParts) {
+            if (! $urlParts) {
```
dhensby, line 34:
```suggestion
            if (!$urlParts) {
```

includes/functions.php
```diff
@@ -38,14 +40,14 @@ function URLtoPath($url, $baseURL = '', $domainBasedCaching = false)
...
-        if ($dirName != '/' && $dirName != '.') {
+        if ($dirName !== '/' && $dirName !== '.') {
             $prefix = $dirName . '/';
         }
         return $prefix . basename($filename);
     }
 }
 
-if (!function_exists('SilverStripe\\StaticPublishQueue\\PathToURL')) {
+if (! function_exists('SilverStripe\\StaticPublishQueue\\PathToURL')) {
```
dhensby, line 50:
```suggestion
if (!function_exists('SilverStripe\\StaticPublishQueue\\PathToURL')) {
```

includes/staticrequesthandler.php
```diff
@@ -2,8 +2,7 @@
 
 namespace SilverStripe\StaticPublishQueue;
 
-return function($cacheDir, $urlMapping = null)
-{
+return function ($cacheDir, $urlMapping = null) {
```
dhensby, line 5:
```suggestion
return function ($cacheDir, $urlMapping = null)
{
```

includes/staticrequesthandler.php
```diff
@@ -18,19 +17,26 @@
...
 
-
     if (is_callable($urlMapping)) {
         $path = $urlMapping($url);
     } else {
         $path = URLtoPath($url);
     }
 
-    if (!$path) {
+    if (! $path) {
```
dhensby, line 26:
```suggestion
    if (!$path) {
```

includes/staticrequesthandler.php
```diff
@@ -18,19 +17,26 @@
...
     }
 
     $cachePath = $cacheDir . DIRECTORY_SEPARATOR . $path;
-
+    $hasHTMLFile = false;
+    $hasGZIPFile = false;
+    if (file_exists($cachePath . '.html')) {
+        $hasHTMLFile = true;
+        $cachePath .= '.html';
+    } elseif (file_exists($cachePath . '.html.gz')) {
```
dhensby, line 36: Should it not check for the gzip first and serve that as a priority? Could both exist at the same time?

includes/staticrequesthandler.php
```diff
@@ -44,26 +50,26 @@
 
     if (file_exists($cachePath . '.php')) {
         $cacheConfig = require $cachePath . '.php';
-    } elseif (!file_exists($cachePath . '.html')) {
+    } elseif (! $hasGZIPFile && ! $hasHTMLFile) {
```
dhensby, line 53:
```suggestion
    } elseif (!$hasGZIPFile && !$hasHTMLFile) {
```

includes/staticrequesthandler.php
```diff
@@ -44,26 +50,26 @@
...
         return false;
     }
     header('X-Cache-Hit: ' . date(\DateTime::COOKIE));
-    if (!empty($cacheConfig['responseCode'])) {
+    if (! empty($cacheConfig['responseCode'])) {
```
dhensby, line 57:
```suggestion
    if (!empty($cacheConfig['responseCode'])) {
```

includes/staticrequesthandler.php
```diff
@@ -44,26 +50,26 @@
...
         header('HTTP/1.1 ' . $cacheConfig['responseCode']);
     }
-    if (!empty($cacheConfig['headers'])) {
+    if (! empty($cacheConfig['headers'])) {
```
dhensby, line 60:
```suggestion
    if (!empty($cacheConfig['headers'])) {
```

includes/staticrequesthandler.php
```diff
@@ -44,26 +50,26 @@
...
         foreach ($cacheConfig['headers'] as $header) {
             header($header, true);
         }
     }
-    if (file_exists($cachePath . '.html')) {
-        $etag = '"' . md5_file($cachePath . '.html') . '"';
-        if (!empty($_SERVER['HTTP_IF_NONE_MATCH']) && $_SERVER['HTTP_IF_NONE_MATCH'] == $etag) {
+    if ($hasHTMLFile || $hasGZIPFile) {
+        $etag = '"' . md5_file($cachePath) . '"';
+        if (! empty($_SERVER['HTTP_IF_NONE_MATCH']) && $_SERVER['HTTP_IF_NONE_MATCH'] === $etag) {
```
dhensby, line 67:
```suggestion
        if (!empty($_SERVER['HTTP_IF_NONE_MATCH']) && $_SERVER['HTTP_IF_NONE_MATCH'] === $etag) {
```

src/Contract/StaticPublisher.php
```diff
@@ -8,16 +8,22 @@ interface StaticPublisher
      * @param string $url
      * @return array A result array
      */
-    public function publishURL($url, $forcePublish = false);
+    public function publishURL(string $url, ?bool $forcePublish = false): array;
```
dhensby, line 11: I think this is a breaking change and we should try to avoid that

src/Contract/StaticPublishingTrigger.php
```diff
@@ -9,22 +9,21 @@
...
 interface StaticPublishingTrigger
 {
-
     /**
      * Provides an SS_List of StaticallyPublishable objects which need to be regenerated.
      *
      * @param array $context An associative array with extra engine-specific information.
      *
-     * @return SS_List
+     * @return array|SS_List
```
dhensby, line 17: You're forcing an array type in the argument declaration but this PHPDoc claims to accept SS_List as well. I think we need to remove the type declarations...

src/Extension/Engine/SiteTreePublishingEngine.php
```diff
@@ -63,7 +63,7 @@ public function setToUpdate($toUpdate)
     }
 
     /**
-     * @param $toDelete
+     * @param mixed $toDelete
```
dhensby, line 66:
```suggestion
     * @param array $toDelete
```

src/Extension/Engine/SiteTreePublishingEngine.php
```diff
@@ -138,7 +138,7 @@ public function collectChanges($context)
     public function flushChanges()
     {
         $queue = QueuedJobService::singleton();
-        if (!empty($this->toUpdate)) {
+        if (! empty($this->toUpdate)) {
```
dhensby, line 141:
```suggestion
        if (!empty($this->toUpdate)) {
```

src/Extension/Engine/SiteTreePublishingEngine.php
```diff
@@ -148,15 +148,15 @@ public function flushChanges()
...
                 ]);
 
                 $queue->queueJob($job);
             }
-            $this->toUpdate = array();
+            $this->toUpdate = [];
         }
 
-        if (!empty($this->toDelete)) {
+        if (! empty($this->toDelete)) {
```
dhensby, line 159:
```suggestion
        if (!empty($this->toDelete)) {
```

src/Job/DeleteStaticCacheJob.php
```diff
@@ -28,12 +28,12 @@ public function process()
...
         $chunkSize = self::config()->get('chunk_size');
         $count = 0;
-        foreach ($this->jobData->URLsToProcess as $url => $priority) {
+        foreach (array_keys($this->jobData->URLsToProcess) as $url) {
             if (++$count > $chunkSize) {
                 break;
             }
             $meta = Publisher::singleton()->purgeURL($url);
-            if (!empty($meta['success'])) {
+            if (! empty($meta['success'])) {
```
dhensby, line 36:
```suggestion
            if (!empty($meta['success'])) {
```

src/Job/GenerateStaticCacheJob.php
```diff
@@ -22,12 +22,12 @@ public function process()
...
         $chunkSize = self::config()->get('chunk_size');
         $count = 0;
-        foreach ($this->jobData->URLsToProcess as $url => $priority) {
+        foreach (array_keys($this->jobData->URLsToProcess) as $url) {
             if (++$count > $chunkSize) {
                 break;
             }
             $meta = Publisher::singleton()->publishURL($url, true);
-            if (!empty($meta['success'])) {
+            if (! empty($meta['success'])) {
```
dhensby, line 30:
```suggestion
            if (!empty($meta['success'])) {
```

src/Publisher.php
```diff
@@ -59,7 +58,7 @@ public function generatePageResponse($url)
             $url = Director::absoluteURL($url);
         }
         $urlParts = parse_url($url);
-        if (!empty($urlParts['query'])) {
+        if (! empty($urlParts['query'])) {
```
dhensby, line 61:
```suggestion
        if (!empty($urlParts['query'])) {
```

src/Publisher.php
```diff
@@ -145,7 +144,7 @@ protected function generatePHPCacheFile($response)
         ];
 
         foreach ($response->getHeaders() as $header => $value) {
-            if (!in_array($header, ['cache-control'])) {
+            if (! in_array($header, ['cache-control'], true)) {
```
dhensby, line 147:
```suggestion
            if (!in_array($header, ['cache-control'], true)) {
```

src/Publisher/FilesystemPublisher.php
```diff
@@ -44,7 +48,7 @@ public function getDestFolder()
     public function setFileExtension($fileExtension)
     {
         $fileExtension = strtolower($fileExtension);
-        if (!in_array($fileExtension, ['html', 'php'])) {
+        if (! in_array($fileExtension, ['html', 'php'], true)) {
```
dhensby, line 51:
```suggestion
        if (!in_array($fileExtension, ['html', 'php'], true)) {
```

src/Publisher/FilesystemPublisher.php
```diff
@@ -19,7 +18,12 @@ class FilesystemPublisher extends Publisher
     /**
      * @var string
      */
-    protected $fileExtension = 'php';
+    protected $fileExtension = 'html';
```
dhensby, line 21: What implication does this change have?

src/Publisher/FilesystemPublisher.php
```diff
@@ -63,11 +67,11 @@ public function getFileExtension()
         return $this->fileExtension;
     }
 
-    public function purgeURL($url)
+    public function purgeURL(string $url): array
     {
-        if (!$url) {
-            user_error("Bad url:" . var_export($url, true), E_USER_WARNING);
-            return;
+        if (! $url) {
```
dhensby, line 72:
```suggestion
        if (!$url) {
```

src/Publisher/FilesystemPublisher.php
```diff
@@ -84,20 +88,27 @@ public function purgeURL($url)
...
      * @param string $url
      * @return array A result array
      */
-    public function publishURL($url, $forcePublish = false)
+    public function publishURL(string $url, ?bool $forcePublish = false): array
     {
-        if (!$url) {
-            user_error("Bad url:" . var_export($url, true), E_USER_WARNING);
-            return;
+        if (! $url) {
```
dhensby, line 104:
```suggestion
        if (!$url) {
```

src/Publisher/FilesystemPublisher.php
```diff
@@ -145,6 +181,10 @@ protected function publishPage($response, $url)
     {
         $success = true;
         if ($path = $this->URLtoPath($url)) {
+            // little hack to make sure we do not include pages with live forms.
+            if (stripos($response->getBody(), 'name="SecurityID"')) {
```
dhensby, line 185: This feels a bit out of scope and is a very fragile implementation because `SecurityID` is configurable. I'd prefer this to be removed from this PR

src/Publisher/FilesystemPublisher.php
```diff
@@ -174,8 +215,25 @@ protected function saveToPath($content, $filePath)
...
+
+        return file_exists($publishPath);
+    }
+
+    protected function compressFile(string $publishPath): string
+    {
+        //we keep the html file for now ... so use second parameter to achieve this.
+        $publishPathGZipped = $publishPath . '.gz';
+        exec('rm ' . $publishPathGZipped . ' -rf');
+        exec('gzip ' . $publishPath);
```
dhensby, line 233: What if gzip is unavailable?

Why can't we gzip with PHP rather than using `exec`? This use of exec makes me a bit worried.

src/Publisher/FilesystemPublisher.php
```diff
@@ -174,8 +215,25 @@ protected function saveToPath($content, $filePath)
...
+        }
+
+        return file_exists($publishPath);
+    }
+
+    protected function compressFile(string $publishPath): string
+    {
+        //we keep the html file for now ... so use second parameter to achieve this.
+        $publishPathGZipped = $publishPath . '.gz';
+        exec('rm ' . $publishPathGZipped . ' -rf');
```
dhensby, line 232: This should *never* be a directory so the use of `-r` should not be necessary and exposes this call to a larger amount of power than is required

src/Publisher/FilesystemPublisher.php
```diff
@@ -174,8 +215,25 @@ protected function saveToPath($content, $filePath)
...
+        exec('gzip ' . $publishPath);
+        exec('chmod 0777 ' . $publishPathGZipped);
```
dhensby, line 234: Why do we want all users to be able to write to this file?

dhensby (changes requested):
There are a fair few linting changes needed. The tests are failing, perhaps because the default publish format has changed from php -> html. The PHP files should be able to deliver the gzipped data just as apache can, which would allow dynamic headers *and* a performance increase - that feature appears to be missing from this PR?

composer.json
sunnysideup: ;-) Sorry, I wanted to use this new version so I changed it temporarily so that I could publish on packagist.

src/Publisher/FilesystemPublisher.php
sunnysideup: Absolutely - DONE

src/Publisher/FilesystemPublisher.php
sunnysideup: Absolutely - DONE

src/Publisher/FilesystemPublisher.php
sunnysideup: reversed!

src/Publisher/FilesystemPublisher.php
sunnysideup: made it optional - I still think this is a very easy way to avoid pages with forms.

sunnysideup (commented):
DONE

sunnysideup:
Ok - so I have done another big cleanup of all the points you raised - awesome.  This is a very optional PR, but if you are keen then I am keen to make it happen!

includes/staticrequesthandler.php
```diff
@@ -18,19 +17,25 @@
...
 
-    if (!$path) {
+    if (! $path) {
         return false;
     }
 
     $cachePath = $cacheDir . DIRECTORY_SEPARATOR . $path;
-
+    $hasCachedFile = true;
+    if (file_exists($cachePath . '.html.gz')) {
```
dhensby, line 32: I think this needs to be a bit more robust. We shouldn't look for the gzipped version if there is no valid `Accept-Encoding` header. We should also always check the html file exists and calculate the ETag from that.

src/Contract/StaticPublishingTrigger.php
dhensby: Sorry, bit of confusion - this is the `@returns` so I think it was fine as it was.

src/Publisher/FilesystemPublisher.php
```diff
@@ -145,16 +166,25 @@ protected function publishPage($response, $url)
     {
         $success = true;
         if ($path = $this->URLtoPath($url)) {
+            $body = $response->getBody();
+            if ($this->Config()->get('lazy_form_recognition')) {
```
dhensby, line 170: wrong case
```suggestion
            if ($this->config()->get('lazy_form_recognition')) {
```

src/Publisher/FilesystemPublisher.php
```diff
@@ -174,8 +204,35 @@ protected function saveToPath($content, $filePath)
...
         Filesystem::makeFolder(dirname($publishPath));
+        $successWithPublish = rename($temporaryPath, $publishPath);
+        if ($successWithPublish) {
+            if (FilesystemPublisher::config()->get('use_gzip_compression')) {
+                $publishPath = $this->compressFile($publishPath);
+            }
+        }
 
-        return rename($temporaryPath, $publishPath);
+        return file_exists($publishPath);
```
dhensby, line 214: I feel like if the gzip compression fails we shouldn't return false - I think it should be an optional step.

src/Publisher/FilesystemPublisher.php
```diff
@@ -174,8 +204,35 @@ protected function saveToPath($content, $filePath)
...
+        $publishPathGZipped = $publishPath . '.gz';
+        // remove
+        unlink($publishPathGZipped);
+        // open a new one
+        $fp = fopen($publishPathGZipped, 'w');
+        // write it
+        fwrite($fp, $gzdata);
+        // close it
+        fclose($fp);
+        @chmod($file, 0664);
```
dhensby, line 233: `$file` is undefined

src/Publisher/FilesystemPublisher.php
```diff
@@ -174,8 +204,35 @@ protected function saveToPath($content, $filePath)
...
+        }
 
-        return rename($temporaryPath, $publishPath);
+        return file_exists($publishPath);
+    }
+
+    protected function compressFile(string $publishPath): string
+    {
+        // read original
+        $data = implode('', file($publishPath));
```
dhensby, line 220: This removes the new lines, doesn't it? perhaps `file_get_contents` would be better?

dhensby (changes requested):
OK - I've had a good look at this now, rebasing and squashing things down from 53 commits to 25, there's still some tidy up to do but this has given me a better understanding of what is going on.

The `use_gzip_compression` needs to be part of `Publisher` and not `FileSystemPublisher`, it should not be a feature solely of the FileSystemPublisher, any implementor of `Publisher` should be able to implement it (optionally).

dhensby:
I'm pulling out the linting / bugs into sepearate PRs so they can be evaluated on their own merit and so that this PR is focused on the GZipping aspects only

dhensby:
OK - latest update.

I've split out all the changes into separate PRs, that allows us to concentrate feedback / work in the right places and to stay focused on getting each feature merged.

I've got the linting changes merged already, which is great news.

So the main feedback for the gzipping is:

1. [ ] We need to make sure the appropriate Accept-Encoding header has been received by the server.
2. [ ] The ETag should be calculated from the HTML file always (if the browser sends a request not accepting gzip and we give them the file uncompressed and then they ask again with the etag and saying they can accept gzip they should still get a 304)
3. [ ] The response needs to include a content-encoding header
4. [ ] We need some tests

dhensby:
Right, this is quite old now and lots of the changes from the (original) PR have been picked out and merged. There's a lot of conflicts to get over which I'm not entirely enthusiastic about resolving... Plus I'm still sceptical about this server-side gzipping, though I'd accept taking it in if there were gzipping happening via a PHP extension and not via a CLI execution..

So in light of this I'm going to close this PR

PR #121, opened 2020-09-16
MINOR: adding hack to exclude pages with forms

dhensby:
Pulled out of #118

### DIFF
--- a/src/Publisher/FilesystemPublisher.php
+++ b/src/Publisher/FilesystemPublisher.php
@@ -4,6 +4,8 @@ namespace SilverStripe\StaticPublishQueue\Publisher;
 
 use SilverStripe\Assets\Filesystem;
 use SilverStripe\Control\HTTPResponse;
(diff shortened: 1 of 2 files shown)


src/Publisher/FilesystemPublisher.php
```diff
@@ -145,11 +147,19 @@ protected function publishPage($response, $url)
     {
         $success = true;
         if ($path = $this->URLtoPath($url)) {
+            $body = $response->getBody();
+            if ($this->config()->get('lazy_form_recognition')) {
+                $id = Config::inst()->get(SecurityToken::class, 'default_name') ?: 'SecurityID';
+                // little hack to make sure we do not include pages with live forms.
+                if (stripos($body, '<input type="hidden" name="' . $id . '"')) {
```
dhensby, line 154:
```suggestion
                if (stripos($body, '<input type="hidden" name="' . $id . '"') !== false) {
```

andrewandante (approved):

PR #122, opened 2020-09-16
More linting fixes

dhensby:


dhensby:
Self merging as these are picked from pr #118

PR #124, opened 2020-09-16
PATCH: add more messages and add default value for variable

dhensby:
plucked from #118


dhensby:
I'm going to close this as I think #130 probably dealt with it

PR #125, opened 2020-10-16
StaticCacheFullBuildJob: Increment current step after chunk

### LINKED ISSUE #103: Job stalls when setting chunk_size lower than the total amount of urls [SS4.3]
I noticed that when setting 
```yaml
SilverStripe\StaticPublishQueue\Job\StaticCacheFullBuildJob:
  chunk_size: 10
```
And running the `StaticCacheFullBuildJob` task the job is marked as `Job stalled after ...` after building the first chunk of URLs.

I believe this happens because of [this if check in queuedjobs](https://github.com/symbiote/silverstripe-queuedjobs/blob/master/src/Services/QueuedJobService.php#L881) which checks the `currentStep` on the job, which is never increased.

Adding `$this->currentStep++;` to the `StaticCacheFullBuildJob` [for example after this line](https://github.com/silverstripe/silverstripe-staticpublishqueue/blob/master/src/Job/StaticCacheFullBuildJob.php#L68) makes the job work as intended.

I can submit a PR for this, but I'm not sure that's the right place to increase the step count, and it would probably also need some sanity checks.

nikrolls:
Chunking doesn't currently work because the current step is not incremented after completing a chunk. QueuedJobs interprents this as a failure and retries the chunk instead; then after a number of retries considers the job stalled.

Incrementing the current step after completing a chunk fixes this and resolve #103.

### DIFF
--- a/src/Job/StaticCacheFullBuildJob.php
+++ b/src/Job/StaticCacheFullBuildJob.php
@@ -66,6 +66,8 @@ class StaticCacheFullBuildJob extends Job
                 unset($this->jobData->URLsToProcess[$url]);
             }
         }
+        $this->currentStep++;
+
         if (empty($this->jobData->URLsToProcess)) {
             $trimSlashes = function ($value) {
                 return trim($value, '/');


dhensby:
I think this may now be fixed via #130

PR #127, opened 2020-12-14
Removes `?stage=Live` from the URL

ssmarco:
Summary:
- Removes `?stage=Live` from the URL which introduces memory leaks

### DIFF
--- a/src/Extension/Publishable/PublishableSiteTree.php
+++ b/src/Extension/Publishable/PublishableSiteTree.php
@@ -7,6 +7,7 @@ use SilverStripe\Control\Director;
 use SilverStripe\ORM\DataExtension;
 use SilverStripe\StaticPublishQueue\Contract\StaticallyPublishable;
 use SilverStripe\StaticPublishQueue\Contract\StaticPublishingTrigger;
+use SilverStripe\Versioned\Versioned;
 
 /**
  * Bare-bones implementation of a publishable page.
@@ -87,9 +88,13 @@ class PublishableSiteTree extends DataExtension implements StaticallyPublishable
 
     /**
      * The only URL belonging to this object is it's own URL.
+     * Explicitly apply Versioned::LIVE to get the absolute URL.
      */
     public function urlsToCache()
     {
-        return [Director::absoluteURL($this->getOwner()->Link()) => 0];
+        return Versioned::withVersionedMode(function () {
+            Versioned::set_reading_mode(Versioned::LIVE);
+            return [Director::absoluteURL($this->getOwner()->Link()) => 0];
+        });
     }
 }


src/Job/StaticCacheFullBuildJob.php
```diff
@@ -61,7 +64,8 @@ public function process()
                 break;
             }
             $meta = Publisher::singleton()->publishURL($url, true);
-            if (!empty($meta['success'])) {
+            // Includes ErrorPage where you can choose your own status codes
+            if (!empty($meta['success']) || $meta['responsecode'] > 400) {
```
andrewandante, line 68: This seems problematic - surely if we get a >400 code response we assume something has gone wrong? Is there a way to check explicitly that the response is coming from an errorpage?

src/Job/GenerateStaticCacheJob.php
```diff
@@ -28,7 +28,6 @@ public function process()
             }
             $meta = Publisher::singleton()->publishURL($url, true);
             if (!empty($meta['success'])) {
-                $this->jobData->ProcessedURLs[$url] = $url;
```
andrewandante, line 31: Why don't we need this anymore?

src/Job/DeleteStaticCacheJob.php
```diff
@@ -34,7 +34,6 @@ public function process()
             }
             $meta = Publisher::singleton()->purgeURL($url);
             if (!empty($meta['success'])) {
-                $this->jobData->ProcessedURLs[$url] = $url;
```
andrewandante, line 37: Same here, why don't we need this?

src/Extension/Publishable/PublishableSiteTree.php
```diff
@@ -90,6 +90,27 @@ public function objectsToDelete($context)
...
     {
-        return [Director::absoluteURL($this->getOwner()->Link()) => 0];
+        $link = $this->getOwner()->Link();
+        if (strpos($link, '?') === false) {
+            return [Director::absoluteURL($link) => 0];
+        }
+
+        // Find querystrings
+        $queryargs = [];
+        if (strpos($link, '?') !== false) {
```
andrewandante, line 100: You don't need this check, you'll only be here if it's true

src/Extension/Publishable/PublishableSiteTree.php
```diff
@@ -90,6 +90,27 @@ public function objectsToDelete($context)
...
+            list($link, $suffix) = explode('?', $link, 2);
+            parse_str($suffix, $localargs);
+            if (!empty($localargs['stage']) && $localargs['stage'] === 'Live') {
+                unset($localargs['stage']);
+            }
+            $queryargs = array_merge($queryargs, $localargs);
```
andrewandante, line 106: You'll also never need this, just use `$localargs` or rename `$localargs`

src/Job/DeleteStaticCacheJob.php
ssmarco: @andrewandante This is a bit controversial but below are my points:
1. This is a single job for a single page, it is enough to know the that processed url is the page itself
2. There isn't any unit tests that requires this, hence removing this code still makes the test passing.
3. There is no evidence that after setting this array that values in it are evaluated for any purpose.
4. Even the return function does not care about it hence this appears to me that we are just using memory allocation for something that is not used in other places to verify or check with.

src/Job/StaticCacheFullBuildJob.php
ssmarco: Good point, I have to dig deeper from this point. With this extra logic, you will end up in a loop thus exhausting memory and mark as broken job.

src/Extension/Publishable/PublishableSiteTree.php
ssmarco: Nice catch 👍

src/Job/GenerateStaticCacheJob.php
ssmarco: https://github.com/silverstripe/silverstripe-staticpublishqueue/pull/127/files#r542996009

src/Extension/Publishable/PublishableSiteTree.php
```diff
@@ -90,6 +90,23 @@ public function objectsToDelete($context)
      */
     public function urlsToCache()
     {
-        return [Director::absoluteURL($this->getOwner()->Link()) => 0];
+        $link = $this->getOwner()->Link();
+        if (strpos($link, '?') === false) {
```
emteknetnz, line 94: Use `parse_url($url, PHP_URL_QUERY)` to extract the query string rather than doing it yourself - https://www.php.net/manual/en/function.parse-url.php

includes/functions.php
```diff
@@ -12,7 +12,13 @@ function URLtoPath($url, $baseURL = '', $domainBasedCaching = false)
 
         // query strings are not yet supported so we need to bail is there is one present
```
emteknetnz, line 13: modify this comment so that:
a) we still say that query strings are not supported
b) however we so still remove stage=live from them

includes/functions.php
```diff
@@ -12,7 +12,13 @@ function URLtoPath($url, $baseURL = '', $domainBasedCaching = false)
...
         if (!empty($urlParts['query'])) {
-            return;
+            parse_str($urlParts['query'], $queryParts);
+            if (!empty($queryParts['stage']) && $queryParts['stage'] === 'Live') {
+                unset($queryParts['stage']);
+            }
+            if (!empty($queryParts)) {
```
emteknetnz, line 19: what's this check for?

src/Job/DeleteStaticCacheJob.php
emteknetnz: I don't get why we're removing this.  If it's doing no harm then I'd prefer to keep this in.  There may be something it's used for that we haven't considered.

src/Job/StaticCacheFullBuildJob.php
emteknetnz: Is there a way to check that page type here to see if it's an ErrorPage instead?

emteknetnz (changes requested):

src/Job/DeleteStaticCacheJob.php
dhensby: Isn't the job data needed for the actual job interface in the CMS so users can see what's been going on. I think it needs to stay.

dhensby:
@ssmarco do you have any interest in continuing with this? It's been a while since it was opened and feedback was given.

ssmarco:
@dhensby I totally forgot about this already. Will have a look this weekend.

includes/functions.php
dhensby: This section is covered in #120 so I think it should be removed

src/Job/StaticCacheFullBuildJob.php
ssmarco: Reverting, this fix is not needed anymore.

src/Job/GenerateStaticCacheJob.php
ssmarco: Reverted

src/Job/DeleteStaticCacheJob.php
ssmarco: Reverted

src/Extension/Publishable/PublishableSiteTree.php
ssmarco: Thanks for the tip, used a different approach.

includes/functions.php
ssmarco: Reverting, this fix is not needed anymore.

ssmarco:
@dhensby Have updated my PR here down to one file after rebasing with latest changes.

ssmarco:
@emteknetnz Have updated my PR here down to one file after rebasing with latest changes.

ssmarco:
**Before**
image

**After**
image

ssmarco:
@andrewandante Have updated my PR here down to one file after rebasing with latest changes.

dhensby:
Hmm, if the problem is that versioned is adding the live stage param, is that because versioned thinks we're in draft mode but trying to get a link to live? Can we just wrap this call in `Versioned:: withVersionedMode()` instead? I'd be worried we might break some other hooks or integrations otherwise.

src/Extension/Publishable/PublishableSiteTree.php
```diff
@@ -87,9 +88,13 @@ public function objectsToDelete($context)
...
      * The only URL belonging to this object is it's own URL.
+     * Instead of Link() which calls VersionedStateExtension::updateLink and appends stage=Live
      */
     public function urlsToCache()
     {
-        return [Director::absoluteURL($this->getOwner()->Link()) => 0];
+        $relativeLink = $this->getOwner()->RelativeLink();
+        $link =  Controller::join_links(Director::baseURL(), $relativeLink);
+
+        return [Director::absoluteURL($link) => 0];
```
dhensby, line 98:
```suggestion
        return Versioned::withVersionedMode(function () {
            Versioned::set_reading_mode(Versioned::get_default_reading_mode());
            return [Director::absoluteURL($this->getOwner()->Link()) => 0];
        });
```

dhensby: Something like this?

ssmarco: Although I know that query string are not supported, but the fix does not have to know whether there is versioning or not.
In this example, it clearly couples with versioning in mind.

Let me try the code and see if it will still work as expected.

ssmarco: I still get `[2021-05-21 21:44:56] error-log.ALERT: Fatal Error (E_ERROR): Maximum execution time of 30 seconds exceeded {"code":1,"message":"Maximum execution time of 30 seconds exceeded","file":"/var/www/staticpublish/vendor/silverstripe/framework/src/Core/Injector/InjectionCreator.php","line":17,"trace":null} []
`

ssmarco: @dhensby It still appends `?stage=Live` to all pages as shown in my previous screenshot which I think versioned still knows that it is still in Live mode.

Another problem I see is that if there is a query string like that, then `Publisher::generatePageResponse($url)` is going to redirect to login page since this is coming from a different session.

dhensby:
I'm going to do a small amount of investigating this locally this weekend.

dhensby:
@ssmarco I can't seem to replicate this locally. When I checkout SS 4.x-dev and install silverstripe/staticpublishqueue:5.0, I don't see the `?stage=Live` URL appended to the URLs to cache and I don't get an error.

Do you have any code that can help me replicate this?

ssmarco:
Hi @dhensby Also did the same thing. SS4.7 and staticpublishqueue:5
To replicate the error:
- Run `dev/tasks/SilverStripe-StaticPublishQueue-Task-StaticCacheFullBuildTask`
- Run the queue job manually in the CMS.

There are two different outcomes
1. The folder public/cache is empty. This is because 
2. Fatal Error with maximum execution time.

src/Extension/Publishable/PublishableSiteTree.php
mfendeksilverstripe: Hey @ssmarco , can you please check what user is used when running your job? You can check this by looking at `RunAsMemberID` DB column. I think we had a similar issue on some other project. Static cache jobs should run without a user to force correct stage (`RunAsMemberID` = 0).

ssmarco: Hi @mfendeksilverstripe, I saw `RunAsID` in `QueuedJobDescriptor` table and all values are zero

mfendeksilverstripe: That's the correct value so this is indeed an issue that needs to be fixed. I had a look at one of the projects where we use this module and this is the current solution:

```
// We never want to have stage=Stage present in our cached URLs. This will safely remove "stage" params, but
// keep any others. It doesn't matter where in the string "stage=" exists.
$url = preg_replace('/([?&])stage=[^&]+(&|$)/', '$1', $url);
// Trim any trailing "?" or "&".
$url = $string = rtrim($url, '&');
$url = $string = rtrim($url, '?');
```
I added some [extensibility improvements](https://github.com/silverstripe/silverstripe-staticpublishqueue/pull/132/files#diff-6552cf24d9678ff8dfa09b29e7d6b2437c951075d96ffd6dbe558e827234f52dR17) in one of my PRs so maybe fix for this can be included in the URLService..

PR #131, opened 2021-05-23
Fix docs for SilverStripe-StaticPublishQueue-Task-StaticCacheFullBuildTask

### LINKED ISSUE #114: Documentation for command to run StaticCacheFullBuildTask incorrect
Hello, I am trying to implement this module on a site that's using Silverstripe 4.4

The documentation says that this is the way to run the task that intitually builds the cache:

`./path/to/sake dev/tasks/StaticCacheFullBuildTask`

However this gives me errors about the task not existing. I tried running `./vendor/bin/sake dev/tasks` to get a list of all tasks... and after grepping for `Static` I found it! The command appears totally different here:

` * SilverStripe\StaticPublishQueue\Task\StaticCacheFullBuildTask: sake dev/tasks/SilverStripe-StaticPublishQueue-Task-StaticCacheFullBuildTask`

To actually get it to run via CLI I have to do:

`./vendor/bin/sake dev/tasks/SilverStripe-StaticPublishQueue-Task-StaticCacheFullBuildTask`

Let me know if I'm doing something wrong or this is some info in the docs that should be updated.

andrewandante:
Fixes #114

### DIFF
--- a/docs/en/building_the_cache.md
+++ b/docs/en/building_the_cache.md
@@ -4,19 +4,19 @@ There are two main ways the cache is built:
 
 ## OnAfterPublish Hook
 
-This calls `urlsToCache()` on a page after it is published. By default, it will cache 
-only its own URL and the ones of its parents. These URLs are added to a 
+This calls `urlsToCache()` on a page after it is published. By default, it will cache
+only its own URL and the ones of its parents. These URLs are added to a
 `GenerateStaticCacheJob` to be executed on the next run.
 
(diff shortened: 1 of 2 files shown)


dhensby:
Is this not going to break existing consumers that use the old link?

andrewandante:
Yeah you're right, I thought it acted like an alias but it's a full-on overwrite. Have fixed the docs instead.

madmatt:
I'm gonna merge this, I'm not really a maintainer of the project but it's a docs-only change and I know it works for me too.

PR #132, opened 2021-05-28
NEW: Static cache batching improvements.

mfendeksilverstripe:
# Static cache batching improvements / fixes

This is yet another extensibility bundle based on feedback from several projects with some light bugfixes sprinkled in.

## SiteTree re-organization fix

The condition did not properly match sometimes due to the type comparison which ended up causing unwanted purge static cache jobs.

## URL bundling

Bundling of URLs into Jobs is now handled by customisable service.

## Job batch size

It's now possible to customise how many URLs are bundled into a single job. The default setting is 0 (put all URLs into a single job). This allows larger websites to manage the batch size based on their infrastructure limits.

## Explicit hydrate method for jobs

This is just a small touch to enforce proper data when creating a new job (no functional impact).

### DIFF
--- a/src/Extension/Engine/SiteTreePublishingEngine.php
+++ b/src/Extension/Engine/SiteTreePublishingEngine.php
@@ -2,14 +2,17 @@
 
 namespace SilverStripe\StaticPublishQueue\Extension\Engine;
 
+use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\CMS\Model\SiteTreeExtension;
 use SilverStripe\Core\Environment;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Resettable;
+use SilverStripe\ORM\ValidationException;
 use SilverStripe\StaticPublishQueue\Contract\StaticPublishingTrigger;
 use SilverStripe\StaticPublishQueue\Extension\Publishable\PublishableSiteTree;
 use SilverStripe\StaticPublishQueue\Job\DeleteStaticCacheJob;
 use SilverStripe\StaticPublishQueue\Job\GenerateStaticCacheJob;
+use SilverStripe\StaticPublishQueue\Service\UrlBundleInterface;
 use Symbiote\QueuedJobs\Services\QueuedJobService;
 
 /**
@@ -99,15 +102,16 @@ class SiteTreePublishingEngine extends SiteTreeExtension implements Resettable
     }
 
     /**
-     * @param \SilverStripe\CMS\Model\SiteTree|null $original
+     * @param SiteTree|SiteTreePublishingEngine|null $original
+     * @throws ValidationException
      */
     public function onAfterPublishRecursive(&$original)
     {
-        // if the site tree has been "reorganised" (ie: the parentID has changed)
-        // then this is eht equivalent of an unpublish and publish as far as the
+        // If the site tree has been "reorganised" (ie: the parentID has changed)
+        // then this is the equivalent of an un-publish and publish as far as the
         // static publisher is concerned
         if ($original && (
-                $original->ParentID !== $this->getOwner()->ParentID
+            (int) $original->ParentID !== (int) $this->getOwner()->ParentID
                 || $original->URLSegment !== $this->getOwner()->URLSegment
             )
         ) {
@@ -132,6 +136,9 @@ class SiteTreePublishingEngine extends SiteTreeExtension implements Resettable
         $this->collectChanges($context);
     }
 
+    /**
+     * @throws ValidationException
+     */
     public function onAfterUnpublish()
     {
         $this->flushChanges();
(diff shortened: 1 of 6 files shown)


dawb:
Has anyone had time to review this and whether it can be released? This would be a great feature for larger websites using the module.

src/Extension/Engine/SiteTreePublishingEngine.php
```diff
@@ -99,15 +102,16 @@ public function setToDelete($toDelete)
...
     public function onAfterPublishRecursive(&$original)
     {
-        // if the site tree has been "reorganised" (ie: the parentID has changed)
-        // then this is eht equivalent of an unpublish and publish as far as the
+        // If the site tree has been "reorganised" (ie: the parentID has changed)
+        // then this is the equivalent of an un-publish and publish as far as the
         // static publisher is concerned
         if ($original && (
-                $original->ParentID !== $this->getOwner()->ParentID
+            (int) $original->ParentID !== (int) $this->getOwner()->ParentID
```
dhensby, line 114: This change seems unnecessary

src/Extension/Engine/SiteTreePublishingEngine.php
```diff
@@ -99,15 +102,16 @@ public function setToDelete($toDelete)
     }
 
     /**
-     * @param \SilverStripe\CMS\Model\SiteTree|null $original
+     * @param SiteTree|SiteTreePublishingEngine|null $original
+     * @throws ValidationException
```
dhensby, line 106: I'm not a PHPDoc expert, but this seems tautologous as it's not this method that throws it's a method that is called within it and I'd have thought that the IDEs/docs generators should be keeping track of that rather than having every single method document the exceptions that methods it calls (and then that those call and so on).

src/Extension/Engine/SiteTreePublishingEngine.php
```diff
@@ -132,6 +136,9 @@ public function onBeforeUnpublish()
         $this->collectChanges($context);
     }
 
+    /**
+     * @throws ValidationException
+     */
```
dhensby, line 141: as above, this seems redundant

src/Extension/Engine/SiteTreePublishingEngine.php
```diff
@@ -160,44 +167,49 @@ public function collectChanges($context)
 
     /**
      * Execute URL deletions, enqueue URL updates.
+     * @throws ValidationException
```
dhensby, line 170: as above, this seems redundant

src/Extension/Engine/SiteTreePublishingEngine.php
```diff
@@ -160,44 +167,49 @@ public function collectChanges($context)
...
      */
     public function flushChanges()
     {
-        $queue = static::$queueService ?? QueuedJobService::singleton();
+        $queueService = static::$queueService ?? QueuedJobService::singleton();
 
-        if (!empty($this->toUpdate)) {
-            foreach ($this->toUpdate as $queueItem) {
-                $job = Injector::inst()->create(GenerateStaticCacheJob::class);
+        if (count($this->toUpdate) > 0) {
```
dhensby, line 176: Why has this changed from `!empty()` to `count() > 0`. The latter is going to be less efficient, there's no point calculating the total length of an array just to assert it's greater than 0, this is what `empty()` is for

src/Extension/Engine/SiteTreePublishingEngine.php
```diff
@@ -160,44 +167,49 @@ public function collectChanges($context)
...
+                $queueService->queueJob($job);
             }
+
             $this->toUpdate = [];
         }
 
-        if (!empty($this->toDelete)) {
-            foreach ($this->toDelete as $queueItem) {
-                $job = Injector::inst()->create(DeleteStaticCacheJob::class);
+        if (count($this->toDelete) > 0) {
```
dhensby, line 196: As above, this is a regressive change

src/Job.php
```diff
@@ -41,6 +41,45 @@ abstract class Job extends AbstractQueuedJob
...
     private static $chunk_size = 200;
 
+    /**
+     * Number of URLs per job allows you to split work into multiple smaller jobs instead of having one large job
+     * this is useful if you're running a queue setup will parallel processing or if you have too many URLs in general
+     * if this number is too high you're limiting the parallel processing opportunity
+     * if this number is too low you're using your resources inefficiently
+     * as every job processing has a fixed overhead which adds up if there are too many jobs
+     *
+     * in case you project is complex and you are struggling to find the correct number
```
dhensby, line 51:
```suggestion
     * in case your project is complex and you are struggling to find the correct number
```

src/Job.php
```diff
@@ -41,6 +41,45 @@ abstract class Job extends AbstractQueuedJob
...
+     * it's possible to move this value to a CMS setting and adjust as needed without the need of changing the code
```
dhensby, line 52:
```suggestion
     * it's possible to move this value to a CMS setting and adjust as needed without the need to change the code
```

src/Job.php
```diff
@@ -41,6 +41,45 @@ abstract class Job extends AbstractQueuedJob
...
+     * use @see Job::getUrlsPerJob() to override the value lookup
+     * you can subclass your jobs and implement your own getUrlsPerJob() method which will look into CMS setting
```
dhensby, line 54:
```suggestion
     * you can subclass your jobs and implement your own getUrlsPerJob() method which can read from the CMS setting
```

src/Job.php
```diff
@@ -41,6 +41,45 @@ abstract class Job extends AbstractQueuedJob
...
+     *
+     * batching capability can be disabled if urls per job is set to 0
+     * in such case, all URLs will be put into one job
```
dhensby, line 57:
```suggestion
     * in this case, all URLs will be processed in a single job
```

src/Service/UrlBundleService.php
```diff
@@ -0,0 +1,129 @@
...
+     * @return array
+     */
+    protected function assignPriorityToUrls(array $urls): array
+    {
+        $priority = 0;
+        $priorityUrls = [];
+
+        foreach ($urls as $url) {
+            $priorityUrls[$url] = $priority;
+            $priority += 1;
```
dhensby, line 124:
```suggestion
            ++$priority;
```

dhensby (changes requested):
Generally a nice approach.

A few stylistic issues / syntax regressions that should be reverted.

Also, I note the comment in `SiteTreePublishingEngine.php` has been updated to use sentence case but the comment in `Job.php` for `$urls_per_job ` static doesn't follow that style at all. It would be good to keep commenting style consistent.

src/Extension/Engine/SiteTreePublishingEngine.php
mfendeksilverstripe: This is just a different linting setup. My team generally avoid using empty but I'm ok with keeping it here as the module linting setup doesn't prevent using empty.

mfendeksilverstripe:
@dhensby I've pushed up new changes as requested, please review.

src/Extension/Engine/SiteTreePublishingEngine.php
dhensby: From a bit of cursory research online, count takes 2× as long, even on empty arrays, so using count is regressive in terms of performance and so should never be used simply to check if arrays have elements or not.

dhensby (approved):

dawb:
@dhensby would you mind adding a tag for this release?

dhensby:
@dawb 5.1.0

PR #136, opened 2022-05-10
BUG: Page parentage update fix.

mfendeksilverstripe:
## SiteTree re-organization fix

The condition did not properly match sometimes due to the type comparison which ended up causing unwanted purge static cache jobs.

Workaround for https://github.com/silverstripe/silverstripe-versioned/issues/360

Split off from https://github.com/silverstripe/silverstripe-staticpublishqueue/pull/132

### DIFF
--- a/src/Extension/Engine/SiteTreePublishingEngine.php
+++ b/src/Extension/Engine/SiteTreePublishingEngine.php
@@ -110,7 +110,7 @@ class SiteTreePublishingEngine extends SiteTreeExtension implements Resettable
         // then this is the equivalent of an un-publish and publish as far as the
         // static publisher is concerned
         if ($original && (
-            $original->ParentID !== $this->getOwner()->ParentID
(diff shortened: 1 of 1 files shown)


dhensby:
@mfendeksilverstripe can you make sure you open PRs from forks in the future, please.

As for this change, we might as well use loose compactors rather than explicit casting in this case? Or is there some benefit to explicitly casting as ints? As a personal peev, I dislike the assumption that DB IDs should be ints because it really ties our hands in the future

michalkleiner:
I'd second Dan's opinion here, allowing string IDs in future to support e.g. UUIDs instead of sequential numeric IDs will need to be a thing, so using loose comparison here would be OK from my perspective, or, if we have to have a strict comparison, going the other way and comparing it as strings.

mfendeksilverstripe:
Switched to loose comparison as per your recommendation @michalkleiner .

cc @dhensby

dhensby (approved):

PR #137, opened 2022-05-12
NEW: Extension points for FilesystemPublisher.

mfendeksilverstripe:
# NEW: Extension points for FilesystemPublisher.

* Extension points added to `generatePageResponse()` to allow better extensibility

### DIFF
--- a/src/Publisher/FilesystemPublisher.php
+++ b/src/Publisher/FilesystemPublisher.php
@@ -109,7 +109,11 @@ class FilesystemPublisher extends Publisher
             return;
         }
         $success = false;
+
+        $this->extend('onBeforeGeneratePageResponse', $url);
         $response = $this->generatePageResponse($url);
+        $this->extend('onAfterGeneratePageResponse', $url, $response);
+
         $statusCode = $response->getStatusCode();
         $doPublish = ($forcePublish && $this->getFileExtension() === 'php') || $statusCode < 400;
 
--- a/src/Publisher.php
+++ b/src/Publisher.php
@@ -10,6 +10,7 @@ use SilverStripe\Control\HTTPResponse_Exception;
 use SilverStripe\Core\Config\Configurable;
 use SilverStripe\Core\CoreKernel;
 use SilverStripe\Core\Environment;
+use SilverStripe\Core\Extensible;
 use SilverStripe\Core\Injector\Injectable;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\FieldType\DBDatetime;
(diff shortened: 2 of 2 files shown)


dhensby:
The idea behind the publisher classes is that developers can easily implement their own and add their own logic easily. There shouldn't really be a need for extension hooks with the current implementation.

What is the scenario where this is needed?

mfendeksilverstripe:
@dhensby My team found useful to have the ability to inject some custom behaviour via these extension points. For example we have in-memory cache which needs to be flushed after each call of `generatePageResponse()`. It's not something that warrants the implementation of a new file system publisher as we're not altering any core behaviour but instead a simple extension can accomodate this requirement.

Also, we found useful to reset some core states such as controller stack something like this:

```php
try {
    $controllers = [];

    // pop all current application's controllers to simulate cold boot
    while (Controller::has_curr()) {
        $controller = Controller::curr();
        $controllers[]= $controller;
        $controller->popCurrent();
    }

    return parent::generatePageResponse($url);
} finally {
    // flush all in-memory cached data

    foreach (array_reverse($controllers) as $controller) {
        // reapply previous controllers
        $controller->pushCurrent();
    }
}
```

src/Publisher/FilesystemPublisher.php
```diff
@@ -68,7 +68,7 @@ public function purgeURL($url)
     {
         if (!$url) {
             user_error('Bad url:' . var_export($url, true), E_USER_WARNING);
-            return;
+            return [];
```
ChloeSartorelli: Why is this change here? It seems out of scope for adding extension points. If there is a valid reason to change what is returned here, please create a separate PR for this purpose which can be reviewed and merged on its own merits.
```suggestion
```

src/Publisher/FilesystemPublisher.php
```diff
@@ -101,10 +101,14 @@ public function publishURL($url, $forcePublish = false)
...
```
ChloeSartorelli:
```suggestion
```

GuySartorelli (changes requested):
Your use case seems sensible to me. Please retarget this PR to the `6` branch and remove the unrelated changes.

mfendeksilverstripe:
I've pushed up changes as per your request @GuySartorelli , please review.

GuySartorelli:
You swapped the target branch _after_ resetting commits, which means the new CI isn't able to run. Can you please force push so that the CI runs correctly?

mfendeksilverstripe:
Done @GuySartorelli

GuySartorelli (approved):
Changes look good to me, will merge when CI goes green.

PR #139, opened 2022-06-08
PHP8 support. PHPUnit 9. Update Travis. Update test config

### LINKED ISSUE #135: PHP8 Support
Using php8 on Ubuntu 20.04 I can install `^4` but not `^5` or `dev-master` which is a little odd (a newer module version not supporting a newer version of PHP that an older module version does).

`composer.json` is locked to `^7.1` for some reason. Is there any WIP for php8 support

### LINKED ISSUE #138: 5.1: SapphireTestState config forces you to prefer-source for development/testing
A new configuration has been added in `5.1`.

```yml
  SilverStripe\Dev\State\SapphireTestState:
    properties:
      States:
        staticPublisherState: '%$SilverStripe\StaticPublishQueue\Dev\StaticPublisherState'
```

```php
public function setUp(SapphireTest $test)
{
    // Prepare queued jobs related functionality for unit test run
    // Disable queue logging
    Injector::inst()->registerService(new QueuedJobsTest_Handler(), QueuedJobHandler::class);

    // Disable special actions of the queue service
    Config::modify()->set(QueuedJobService::class, 'use_shutdown_function', false);

    // It seems Injector doesn't cover all cases so we force-inject a test service which is suitable for unit tests
    Injector::inst()->registerService(new QueuedJobsTestService(), QueuedJobService::class);
    SiteTreePublishingEngine::setQueueService(QueuedJobService::singleton());
}
```

`QueuedJobsTest_Handler` is a `test` class from the QueuedJobs module, and `QueuedJobsTestService` is a `test` class from this module. This means they're not available unless you `composer install --prefer-source`.

Is this the expected behaviour?

chrispenny:
Fixes #135
Fixes #138

## PHP Unit 9

Not too much had to update here. Just the usual missing `: void` declarations, and I needed to update some deprecated methods and `assertContains` to the new `assetStringContainsString`.

## Travis update

I'm not 100% across all of the settings that used to be there (in particular the cache dir settings). I've really just grabbed what is the latest "template" from some other Silverstripe modules. 🤞 

## Updated yml config

- [x] Manually tested that a project without `--prefer-source` can successfully run its own unit tests without these missing classes throwing an error (as described in #138)

## Practical test

- [x] Manually test that project purges URLs when unpublished
- [x] Manually test that project builds URLs when published
- [x] Manually test that project rebuilds parent URLs when child is published or unpublished

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,57 +1,25 @@
-language: php
+version: ~> 1.0
 
-dist: xenial
+import:
+  - silverstripe/silverstripe-travis-shared:config/provision/standard.yml
 
-services:
-  - mysql
-  - postgresql
-
-cache:
-  directories:
-    - $HOME/.composer/cache/files
-
-addons:
-  apt:
-    packages:
-      - tidy
-
-matrix:
+jobs:
   fast_finish: true
   include:
-    - php: 7.1
-      env: DB=MYSQL RECIPE_VERSION=4.6.x-dev PHPUNIT_TEST=1 PHPCS_TEST=1
-    - php: 7.2
-      env: DB=PGSQL RECIPE_VERSION=4.6.x-dev PHPUNIT_TEST=1
-    - php: 7.3
-      env: DB=MYSQL RECIPE_VERSION=4.6.x-dev PDO=1 PHPUNIT_TEST=1
     - php: 7.4
-      env: DB=MYSQL RECIPE_VERSION=4.6.x-dev PHPUNIT_TEST=1
-      addons:
-        apt:
-          packages:
-            - libonig-dev
-
-before_script:
-  # Extra $PATH
-  - export PATH=~/.composer/vendor/bin:$PATH
-
-  # Init PHP
-  - phpenv rehash
-  - phpenv config-rm xdebug.ini
-  - export PATH=~/.composer/vendor/bin:$PATH
-  - echo 'memory_limit = 2G' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
-
-  # Install composer dependencies
-  - composer validate
-  - composer require --no-update silverstripe/recipe-cms:$RECIPE_VERSION silverstripe/recipe-testing:^1
-  # Fix for running phpunit 5 on php 7.4+
-  - composer require --no-update sminnee/phpunit-mock-objects:^3
-  - if [[ $DB == PGSQL ]]; then composer require --no-update silverstripe/postgresql:^2; fi
-  - composer install --prefer-source --no-interaction --no-progress --no-suggest --optimize-autoloader --verbose --profile
-
-script:
-  - if [[ $PHPUNIT_TEST ]]; then vendor/bin/phpunit tests/php/ flush=1; fi
-  - if [[ $PHPCS_TEST ]]; then composer run-script lint; fi
-
-after_success:
-  - if [[ $PHPUNIT_COVERAGE_TEST ]]; then bash <(curl -s https://codecov.io/bash) -f coverage.xml; fi
+      env:
+        - DB=MYSQL
+        - REQUIRE_INSTALLER="4.10.x@dev"
+        - PHPCS_TEST=1
+        - PHPUNIT_TEST=1
+        - PHPUNIT_COVERAGE_TEST=1
+    - php: 8.0
+      env:
+        - DB=MYSQL
+        - REQUIRE_INSTALLER="4.10.x@dev"
+        - PHPUNIT_TEST=1
+    - php: 8.1
+      env:
+        - DB=MYSQL
+        - REQUIRE_INSTALLER="4.10.x@dev"
+        - PHPUNIT_TEST=1
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,53 +1,25 @@
-language: php
+version: ~> 1.0
 
-dist: xenial
+import:
+  - silverstripe/silverstripe-travis-shared:config/provision/standard.yml
 
-services:
-  - mysql
-  - postgresql
-
-cache:
-  directories:
-    - $HOME/.composer/cache/files
-
-addons:
-  apt:
-    packages:
-      - tidy
-
-matrix:
+jobs:
   fast_finish: true
   include:
     - php: 7.4
-      env: DB=MYSQL RECIPE_VERSION=4.10.x-dev PHPUNIT_TEST=1 PHPCS_TEST=1
+      env:
+        - DB=MYSQL
+        - REQUIRE_INSTALLER="4.10.x@dev"
+        - PHPCS_TEST=1
+        - PHPUNIT_TEST=1
+        - PHPUNIT_COVERAGE_TEST=1
     - php: 8.0
-      env: DB=MYSQL RECIPE_VERSION=4.10.x-dev PHPUNIT_TEST=1
+      env:
+        - DB=MYSQL
+        - REQUIRE_INSTALLER="4.10.x@dev"
+        - PHPUNIT_TEST=1
     - php: 8.1
-      env: DB=MYSQL RECIPE_VERSION=4.10.x-dev PHPUNIT_TEST=1
-      addons:
-        apt:
-          packages:
-            - libonig-dev
-
-before_script:
-  # Extra $PATH
-  - export PATH=~/.composer/vendor/bin:$PATH
-
-  # Init PHP
-  - phpenv rehash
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -1,57 +1,25 @@
-language: php
```
mfendeksilverstripe, line 1: I recommend trying a very basic version of this config:

```yaml
version: ~> 1.0

env:
  global:
    - COMPOSER_ROOT_VERSION="4.x-dev"
    - REQUIRE_RECIPE="4.x-dev"

import:
  - silverstripe/silverstripe-travis-shared:config/provision/standard-jobs-fixed.yml
```

composer.json
```diff
@@ -18,15 +18,15 @@
...
-        "php": "^7.1",
+        "php": "^7.4 || ^8.0",
         "silverstripe/framework": "^4.6",
         "silverstripe/cms": "^4",
         "silverstripe/config": "^1",
         "symbiote/silverstripe-queuedjobs": "^4.5",
         "silverstripe/versioned": "^1.6"
     },
     "require-dev": {
-        "phpunit/phpunit": "^5.7",
```
mfendeksilverstripe, line 29: I recommend using a testing recipe here instead of explicit modules.
```
"silverstripe/recipe-testing": "^1 || ^2",
```

with that you can remove both `phpunit/phpunit` and `squizlabs/php_codesniffer`

tests/php/Publisher/FilesystemPublisherTest.php
```diff
@@ -193,70 +195,80 @@ public function testOnlyHTML()
...
 
-    public function testPurgeURLAfterSwitchingExtensions()
+    public function testPurgeURLAfterSwitchingExtensions(): void
     {
         $level1 = new StaticPublisherTestPage();
         $level1->URLSegment = 'purge-me';
         $level1->write();
         $level1->publishRecursive();
 
+        $this->logOut();
```
mfendeksilverstripe, line 236: Here is another solution for the case if you specifically want to test the case when a user is logged in.

```php
// Make sure that we retain Cookie_Backend service as @see Publisher::getHTTPApplication()
// creates a fresh Injector instance throwing away any Injector level configuration in the process
Injector::inst()->registerService(new CookieJarMock(), Cookie_Backend::class);
```

```php
<?php

namespace App\Tests\StaticPublishing;

use SilverStripe\Control\CookieJar;
use SilverStripe\Dev\TestOnly;

/**
 * Test only version of @see CookieJar to provide ability to skip cookies output as it may cause unit test failures
 * for tests which execute output rendering such as static cache jobs
 */
class CookieJarMock extends CookieJar implements TestOnly
{
    public function set($name, $value, $expiry = 90, $path = null, $domain = null, $secure = false, $httpOnly = true)
    {
        // No op
    }

    public function forceExpiry($name, $path = null, $domain = null, $secure = false, $httpOnly = true)
    {
        // No op
    }
}
```

mfendeksilverstripe:
@chrispenny We're almost there, please add the following config to the `phpcs.xml.dist` just under the `<description>` tag.

```xml
    <file>src</file>
    <file>tests</file>
```

mfendeksilverstripe:
Here is an example of phpcs file:

https://github.com/silverstripe/silverstripe-versioned-snapshots/blob/1/phpcs.xml.dist

tests/php/Publisher/FilesystemPublisherTest.php
chrispenny: Thanks, @mfendeksilverstripe!

Would I add that `Injector` portion into our existing `StaticPublisherState` (`TestState`)?

chrispenny: Hmmmm, that didn't seem to work. I think I'll just stick with the `logOut()` for now. Adding the Mock can be an improvement if it's needed down the line.

mfendeksilverstripe:
All ticks are green :D Nice one @chrispenny

chrispenny:
Thank you for all of the help, @mfendeksilverstripe!

I've gone through my own regression tests for the project I'm working on, and I'm feeling pretty happy that the existing features that we rely on are still functioning.

mfendeksilverstripe:
Hey @chrispenny , I noticed the following warning in the CI run:

```
Warning:       Your XML configuration validates against a deprecated schema.
Suggestion:    Migrate your XML configuration using "--migrate-configuration"!
```

This is coming from the PHPUnit 9 and we should be able to fix it by updating the `phpunit.xml.dist`. I think the XML file has now a different structure and naming conventions but you should be able to run the automatic update via `phpunit --migrate-configuration` command

chrispenny:
Thanks, @mfendeksilverstripe!

I added allowed plugins (for composer v2), and ran the `migration-configuration` process.

.travis.yml
```diff
@@ -1,57 +1,9 @@
...
+version: ~> 1.0
```
emteknetnz, line 1: Could you replace this with https://github.com/silverstripe/gha-ci - we're replacing travis with this

You may need to do this in a separate PR that we merge before this one

composer.json
```diff
@@ -18,16 +18,15 @@
...
-        "squizlabs/php_codesniffer": "^3"
+        "silverstripe/recipe-testing": "^1 || ^2"
```
emteknetnz, line 29:
```suggestion
        "silverstripe/recipe-testing": "^2"
```

emteknetnz: Though does this even need recipe-testing? Isn't `"phpunit/phpunit": "^9.5"` enough?

phpunit.xml.dist
```diff
@@ -1,13 +1,15 @@
...
+        <include>
+            <directory suffix=".php">src/</directory>
+        </include>
+        <exclude>
+            <directory suffix=".php">tests/</directory>
+        </exclude>
+    </coverage>
     <testsuite name="Default">
         <directory>tests/</directory>
     </testsuite>
```
emteknetnz, line 14:
```suggestion
    <testsuites>
        <testsuite name="Default">
            <directory>tests/</directory>
        </testsuite>
    </testsuites>
```

tests/php/Publisher/FilesystemPublisherTest.php
```diff
@@ -53,62 +53,62 @@ protected function setUp()
...
 
-    public function testUrlToPathWithRelativeUrls()
+    public function testUrlToPathWithRelativeUrls(): void
     {
         $reflection = new \ReflectionClass(FilesystemPublisher::class);
         $urlToPath = $reflection->getMethod('URLtoPath');
         $urlToPath->setAccessible(true);
 
-        $this->assertSame(
+        $this->assertEquals(
```
emteknetnz, line 70: We should be using assertSame, not assertEquals

If there's broken tests, then update the $expected value rather than just using assertEquals

composer.json
```diff
@@ -51,5 +50,12 @@
...
     "prefer-stable": true,
-    "minimum-stability": "dev"
+    "minimum-stability": "dev",
+    "config": {
+        "allow-plugins": {
+            "composer/installers": true,
+            "silverstripe/recipe-plugin": true,
+            "silverstripe/vendor-plugin": true
+        }
+    }
```
emteknetnz, line 60: Please remove the branch-alias from composer.json

I've created new `5` and `5.1` branches from `master` and retargetted this PR to `5`

Please also create a follow up PR to remove the branch alias from the `master` branch

You'll want to check relevant projects that they're requiring this as `^5` rather than `dev-master`

emteknetnz (changes requested):

composer.json
chrispenny: Not sure. I just went with Mojmir's recommendation.

.travis.yml
chrispenny: I've raised #140

Can this happen after this PR?

composer.json
chrispenny: I don't really understand what the consequences of those changes are, but they kinda sound like general house keeping? Does this all need to part of this PHP 8 upgrade PR?

.travis.yml
emteknetnz: yeah after is fine

composer.json
emteknetnz: Preferably now .. we're targeting this at the `5` branch (which should have always existed), so if we keep the existing branch-alias's in the `5` and `master` branches there' a chance that requesting `5.x-dev` will mistakenly fetch from the `master branch` when it shouldn't

emteknetnz (approved):
Happy with changes here

PR #141, opened 2022-06-20
5.2: Fix order of setup() for Full Build Job

chrispenny:
Apologies team, I should have picked this up during my PHP 8 PR testing. Turns out... the project I was testing on has it's own Job called the exact same thing, so I **thought** I was testing this, but I was testing our bespoke one /face-palm.

`parent::setup()` will trigger the following assignment:
```php
$this->totalSteps = count($this->URLsToProcess)
```

So, `$this->URLsToProcess` needs to be set before that happens.

I'm not too sure why this worked before PHP 8 (or maybe it didn't?).

### DIFF
--- a/src/Job/StaticCacheFullBuildJob.php
+++ b/src/Job/StaticCacheFullBuildJob.php
@@ -42,11 +42,12 @@ class StaticCacheFullBuildJob extends Job
 
     public function setup(): void
     {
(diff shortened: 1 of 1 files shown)


mfendeksilverstripe (approved):
LGTM 👍

PR #147, opened 2022-10-05
BUG: Protect against URLsToProcess being falsy

textagroup:
If URLsToProcess is null then a error will be thrown when trying to create a Delete cache job because array_keys expects an array.

[Emergency] Uncaught TypeError: array_keys(): Argument `#1` ($array) must be of type array, null given
POST /admin/queuedjobs/Symbiote-QueuedJobs-DataObjects-QueuedJobDescriptor/EditForm
Line 107 in /var/www/stats/vendor/silverstripe/staticpublishqueue/src/Job.ph

### DIFF
--- a/src/Job.php
+++ b/src/Job.php
@@ -104,6 +104,9 @@ abstract class Job extends AbstractQueuedJob
 
     public function getSignature(): string
     {
+        if (!$this->URLsToProcess) {
+            return md5(static::class);
+        }
         return md5(implode('-', [static::class, implode('-', array_keys($this->URLsToProcess))]));
     }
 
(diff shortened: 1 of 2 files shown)


src/Job.php
```diff
@@ -104,7 +104,8 @@ public function setup(): void
 
     public function getSignature(): string
     {
-        return md5(implode('-', [static::class, implode('-', array_keys($this->URLsToProcess))]));
+        return md5(implode('-', [static::class, implode('-',
+            array_keys((array)$this->URLstoProcess))]));
```
GuySartorelli, line 108:
```suggestion
        if (!$this->URLsToProcess) {
            return md5(static::class);
        }
        return md5(implode('-', [static::class, implode('-', array_keys($this->URLsToProcess))]));
```
If there's nothing to process, then treating it as an array is improper - we should just have a signature that doesn't try to incorporate the URLs in that scenario.

GuySartorelli (changes requested):
Please also include a test where a job is processed with no URLs to process.

GuySartorelli:
The only change I've made is resolving a linting issue

GuySartorelli:
@chrispenny You've been doing a lot of looking into this module recently and I'm not very familiar with it. This PR looks okay but can you please give it a quick glance and let me know if there's anything obviously problematic about this change?

chrispenny:
Sorry team, I've been away on annual leave.

The title of the PR makes it sound like we're going to make sure `URLsToProcess` is cast to an array, but instead it seems we're just checking that it is an array before attempting to loop over it. Either way sounds fine to me.

GuySartorelli (approved):
Looks good to me then. Thanks

PR #148, opened 2022-10-26
feat: add support for immediate flag

wilr:


dhensby:
I think we need a bit more info about what is meant to be going on here, including some docs and tests?

wilr:
@dhensby I think that's overkill given `StaticCacheFullBuildTask` isn't covered by tests at all at this stage and a test wouldn't cover much. I've added docs for running it.

GuySartorelli:
If `$request->getVar('startAfter')` is null, the message "added to the queue for immediate processing" is outputted - which suggests the way to run a job immediately is to not include a "startAfter" time.

How does this PR differ from just not including a "startAfter" time?

PR #149, opened 2022-11-02
FIX added php8.1 compatibility

### LINKED ISSUE #146: staticpublisher queued job fails in some circumstances
Running into problems with SS4.11 / Fluent V6 in combination with the Static Publisher.

Situation:
- Fluent 6 
- SS 4.11
- Running in CLI
- Empty Job queue
- No configuration added for staticpublisher (no urlsToCache() ).

**Run StaticCacheFullBuildTask**  -->  added to the queue for immediate processing
**Run ProcessJobQueueTask** -->  Argument #1 ($value) must be of type Countable|array, null given {"exception":"[object] (TypeError(code: 0): count(): Argument #1 ($value) must be of type Countable|array, null given at /data/site/vendor/silverstripe/staticpublishqueue/src/Job.php:102)"

When I re-publish a page in the CMS,  it gets added to the queue and publishes correctly if I process the queue again.

If I try to run publisher tasks in the CMS jobs UI itself ,  the task runs,  but takes a loooooong time and kills the session. Will have to look at the logs to see what is happening.

satrun77:
Fixes #146

Changes from applying https://github.com/emteknetnz/php81-fixer

### DIFF
--- a/docs/examples/index.php
+++ b/docs/examples/index.php
@@ -22,7 +22,7 @@ $requestHandler = require 'staticrequesthandler.php';
 if ($requestHandler('cache') !== false) {
     die;
 }
-header('X-Cache-Miss: ' . date(\Datetime::COOKIE));
+header('X-Cache-Miss: ' . date(\Datetime::COOKIE ?? ''));
 
 // Build request and detect flush
 $request = HTTPRequestBuilder::createFromEnvironment();
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -8,12 +8,12 @@ if (!function_exists('SilverStripe\\StaticPublishQueue\\URLtoPath')) {
         // parse_url() is not multibyte safe, see https://bugs.php.net/bug.php?id=52923.
         // We assume that the URL hsa been correctly encoded either on storage (for SiteTree->URLSegment),
         // or through URL collection (for controller method names etc.).
-        $urlParts = @parse_url($url);
+        $urlParts = @parse_url($url ?? '');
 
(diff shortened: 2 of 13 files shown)


michalkleiner:
I get the fixer doesn't understand the context of the code and adds a ton of fallback values to methods that expect string etc., but it produces very much a ton of unnecessary modifications for cases where the same variable is used in multiple subsequent calls and the first sanitisation is already enough etc.

My slight worry is that it won't ever get cleaned up and the change is not addressing the actual offending calls that might be one or two of them rather than just blanket applying a fixer, so from that perspective, I'd welcome other views on this.

GuySartorelli:
Personally I think this is okay - we had to do this across core modules when we released 4.11 and I thought it had been done across kitchen sink (including this module) as well at the same time.
This PR basically just does what we should have done at that time.

> My slight worry is that it won't ever get cleaned up and the change is not addressing the actual offending calls that might be one or two of them rather than just blanket applying a fixer

This fix isn't meant to go through and find everything that's calling these methods and passing in null when a string would be preferred (especially since these methods probably don't have typehinting anyway so custom project code could be causing these deprecation warnings as well) - it's retaining the exact same behaviour and just respecting the deprecations in PHP 8.1.

At least that was the original thinking when Steve did the work for 4.11

GuySartorelli:
This hasn't been touched since last year and the consensus seems to be that it's not needed. Closing this PR. If there's more information that indicates the need for it we can look at reopening it.

PR #152, opened 2022-11-23
Added example for trigger use to documentation

lars-lemon8:
Added example regarding the use of StaticPublishingTrigger

### DIFF
--- a/docs/en/basic_configuration.md
+++ b/docs/en/basic_configuration.md
@@ -73,6 +73,34 @@ action. This module provides the following engine:
 * `SiteTreePublishingEngine` - regenerates the static cache files after a page is published/unpublished. This makes use
 of both interfaces as object trigger refresh of other objects.
 
+#### Example:  Triggering a republish of the homepage after an update to a DataObject
+
+In order for the triggers to get picked up and executed, make sure to add the SiteTreePublishingEngine to your DataObject.
+```yaml
+Project\ORM\YourDataObject:
+  extensions:
+    - SilverStripe\StaticPublishQueue\Extension\Engine\SiteTreePublishingEngine
+```
+
+Then implement the StaticPublishingTrigger interface on your DataObject:
+
+```php
+class YourDataObject extends DataObject implements staticPublishingTrigger 
+{
+
+    public function objectsToUpdate($context) 
+    {
+        return HomePage::get()->First();  // return 1 or multiple publishable objects
+    }
+
+     public function objectsToDelete($context)  
+     {
+        // return your publishable objects here
+     }
+}
+
+```
+
 ## Cache information / Debugging
 
 If required, a timestamp can be added to the generated HTML to show when the file was created.  The timestamp is added as an HTML comment immediately before the closing `</html>` tag (on pages which do not include this tag, enabling this functionality will have no effect).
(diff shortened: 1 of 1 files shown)


docs/en/basic_configuration.md
```diff
@@ -73,6 +73,34 @@ action. This module provides the following engine:
...
+```yaml
+Project\ORM\YourDataObject:
+  extensions:
+    - SilverStripe\StaticPublishQueue\Extension\Engine\SiteTreePublishingEngine
+```
+
+Then implement the StaticPublishingTrigger interface on your DataObject:
+
+```php
+class YourDataObject extends DataObject implements staticPublishingTrigger
```
dhensby, line 88: Can you include the import statements in the example code, please.

Also, please use the correct case for the interface (`StaticPublishingTrigger`)

docs/en/basic_configuration.md
```diff
@@ -73,6 +73,34 @@ action. This module provides the following engine:
...
+
+Then implement the StaticPublishingTrigger interface on your DataObject:
+
+```php
+class YourDataObject extends DataObject implements staticPublishingTrigger 
+{
+
+    public function objectsToUpdate($context) 
+    {
+        return HomePage::get()->First();  // return 1 or multiple publishable objects
```
dhensby, line 93: Just for my certainty, this can return an array or a single item?

docs/en/basic_configuration.md
```diff
@@ -73,6 +73,34 @@ action. This module provides the following engine:
...
+    }
+
+     public function objectsToDelete($context)  
+     {
+        // return your publishable objects here
```
dhensby, line 98: maybe more detail here would be useful - eg: if your dataobject has a relationship with other statically published items, those would need to be returned here so their cache is regenerated.

dhensby (changes requested):

docs/en/basic_configuration.md
lars-lemon8: Yes, both.   I updated the comment.

docs/en/basic_configuration.md
lars-lemon8: Done, thanks !

lars-lemon8:
Thanks @dhensby ,  I processed your feedback.

docs/en/basic_configuration.md
```diff
@@ -73,6 +73,36 @@ action. This module provides the following engine:
...
+```yaml
+Project\ORM\YourDataObject:
+  extensions:
+    - SilverStripe\StaticPublishQueue\Extension\Engine\SiteTreePublishingEngine
+```
+
+Then implement the StaticPublishingTrigger interface on your DataObject:
+
+```php
+use SilverStripe\StaticPublishQueue\Contract\StaticPublishingTrigger;
```
dhensby, line 88: You should also have an import for `DataObject` (and potentially `HomePage` too - as it's in the example code, maybe just replace `HomePage` with `Page`?)

dhensby (approved):
Just a minor addition, but I think it's about OK like this

docs/en/basic_configuration.md
lars-lemon8: Agh.. ;-)   

You are right, fixed ! ;-)

dhensby:
Thanks @lars-lemon8! 🎉

Issue #154, opened 2022-12-09
Don't cache redirector pages


emteknetnz:
Redirector pages can point to pages that should be excluded via `urlsToCache()`

~~It may still be desirable to allow caching of redirector pages, so for backwards compatibility provide an opt-in config to disable caching of redirector pages.~~

## Acceptance criteria
- [ ] Static publisher caches the redirector page and not the target.

### PRs
- https://github.com/silverstripe/silverstripe-staticpublishqueue/pull/156

emteknetnz:
Linked PR has been merged

PR #155, opened 2022-12-09
NEW Option to not cache RedirectorPages

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-staticpublishqueue/issues/154

No unit tests because I'm struggling to unit test a `RedirectorPage extends Page` when `Page` doesn't exist in the module, though it does inside a project, and also the assumption is the `Page::urlsToCache()` exists, which again it may or may not depending where the unit test is run.

### DIFF
--- a/src/Job/StaticCacheFullBuildJob.php
+++ b/src/Job/StaticCacheFullBuildJob.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\StaticPublishQueue\Job;
 
+use SilverStripe\CMS\Model\RedirectorPage;
 use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\StaticPublishQueue\Contract\StaticallyPublishable;
 use SilverStripe\StaticPublishQueue\Extension\Publishable\PublishableSiteTree;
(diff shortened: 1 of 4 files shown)


src/Job/StaticCacheFullBuildJob.php
```diff
@@ -80,24 +83,26 @@ public function process(): void
...
         if (count($this->URLsToProcess) === 0) {
-            $trimSlashes = function ($value) {
-                $value = trim($value, '/');
+            $cleanUrl = function ($value) {
 
                 // We want to trim the schema from the beginning as they map to the same place
                 // anyway.
                 $value = ltrim($value, 'http://');
                 $value = ltrim($value, 'https://');
+                $value = preg_replace('#[\?&]stage=' . Versioned::LIVE . '#', '', $value);
```
emteknetnz, line 92: This was a necessary bug fix for me to run locally without it trigger an infinite loop.  It's possibly related to running framework 4.12, though I'm not entirely sure.

src/Publisher.php
```diff
@@ -56,6 +56,11 @@ abstract class Publisher implements StaticPublisher
      */
     private static $add_timestamp = false;
 
+    /**
+     * Cache redirector pages which will cache as the page they're redirected to
+     */
+    private static $cache_redirector_pages = true;
```
emteknetnz, line 62: I'm leaving this a true by default for backwards compatibility, I'd probably prefer it to be false by default though - though that should really require a new major version

dhensby:
I'm a bit confused by this. So when a redirector page is cached it caches its destination and not just the redirect?

That sounds unexpected and a bug that should be fixed rather than adding tight coupling to `RedirectorPage`.

If there's some opinionated caching of pages that are the target of a redirect which are not marked as cachable, then that sounds like a more general problem to fix (what if a page type redirects that's *not* a redirector page; this fix won't help)

GuySartorelli:
Unit tests for this module are run in our standard CI which adds installer which includes the `Page` class. I think you should be able to add a unit test and just have it skip if the `Page` class is missing (i.e. if running tests locally)

emteknetnz:
> So when a redirector page is cached it caches its destination and not just the redirect?

That's correct I believe, though it probably shouldn't

Having a look through the code - staticpublish queue works where it basically does a [fake request](https://github.com/silverstripe/silverstripe-staticpublishqueue/blob/5.2/src/Publisher.php#L114) 

It **should** [save redirects](https://github.com/silverstripe/silverstripe-staticpublishqueue/blob/5.2/src/Publisher/FilesystemPublisher.php#L113) as their own thing - but in my experience it wasn't

emteknetnz:
Closing in favor of https://github.com/silverstripe/silverstripe-staticpublishqueue/pull/156

PR #156, opened 2022-12-11
FIX Cache RedirectorPage as redirect instead of as target

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-staticpublishqueue/issues/154

Using instead of https://github.com/silverstripe/silverstripe-staticpublishqueue/pull/155

Staticpublishqueue does have functionality to [statically cache redirects as their own thing](https://github.com/silverstripe/silverstripe-staticpublishqueue/blob/5.2/src/Publisher/FilesystemPublisher.php#L113) - so I think it's correct to cache RedirectorPages as their own thing, rather than just following the target

### DIFF
--- a/src/Extension/Publishable/PublishableSiteTree.php
+++ b/src/Extension/Publishable/PublishableSiteTree.php
@@ -7,6 +7,7 @@ use SilverStripe\Control\Director;
 use SilverStripe\ORM\DataExtension;
 use SilverStripe\StaticPublishQueue\Contract\StaticallyPublishable;
 use SilverStripe\StaticPublishQueue\Contract\StaticPublishingTrigger;
+use SilverStripe\CMS\Model\RedirectorPage;
 
 /**
  * Bare-bones implementation of a publishable page.
@@ -90,6 +91,14 @@ class PublishableSiteTree extends DataExtension implements StaticallyPublishable
      */
     public function urlsToCache()
     {
-        return [Director::absoluteURL($this->getOwner()->Link()) => 0];
+        $page = $this->getOwner();
+        if ($page instanceof RedirectorPage) {
+            // use RedirectorPage::regularLink() so that it returns the url of the page,
+            // rather than the url of the target of the RedirectorPage
+            $link = $page->regularLink();
+        } else {
+            $link = $page->Link();
+        }
+        return [Director::absoluteURL($link) => 0];
     }
 }
--- a/tests/php/PublishableSiteTreeTest.php
+++ b/tests/php/PublishableSiteTreeTest.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\StaticPublishQueue\Test;
 
+use Page;
 use PHPUnit\Framework\MockObject\Stub\ReturnCallback;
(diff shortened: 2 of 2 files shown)


src/Extension/Publishable/PublishableSiteTree.php
```diff
@@ -90,6 +91,14 @@ public function objectsToDelete($context)
      */
     public function urlsToCache()
     {
-        return [Director::absoluteURL($this->getOwner()->Link()) => 0];
+        $page = $this->getOwner();
+        if ($page instanceof RedirectorPage) {
```
GuySartorelli, line 95: I still feel like explicitly calling out `RedirectorPage` is a bit of an anti-pattern here - unless other redirects are _already_ working correctly, and this problem really is specifically only affecting this page type.

emteknetnz: I think it's fine, RedirectorPage lives in the cms module, and cms is a requirement in composer.json

The reason this fix works is that staticpublish queue will:
- collect a list of **urls** to cache
- make a fake to those urls
- cache the results - if the returned http status code is 300-399, cache as a redirect, else cache as a regular page

If we use RedirectorPage::Link() - it'll return the url of the target page
However, if we use RedirectorPage::regularLink() we get the url of the RedirectorPage - which will return a 301/302 when you normally visit it

GuySartorelli: My issue isn't that `RedirectorPage` might somehow be missing, but rather that it might not be the _only_ way for redirects to be present in a project. It looks like this module is intended to cache _any_ redirects, not just redirects from this page type.

So like I said: If other redirects _do_ still get cached correctly and this issue _is unique_ to the redirector page type, this solution is fine.
But if other redirects _are still not cached correctly_ then I think the argument @dhensby made on the original PR still stands.

emteknetnz: Have confirmed, existing redirect function correct.  It's only RedirectoPage that has the issue.

**MyRedirectPage.php**
```php
<?php

class MyRedirectPage extends Page
{
}
```

**MyRedirectPageController.php**
```php
<?php

use SilverStripe\CMS\Model\SiteTree;

class MyRedirectPageController extends PageController
{
    public function init()
    {
        parent::init();
        $this->redirect(SiteTree::get()->filter(['Title' => 'Contact us'])->first()->Link());
    }
}
```

Creating an instance of this page, publishing it and then running the queued job will will result in a static redirect files created in /public/cache/new-my-redirect-page.html + /public/cache/new-my-redirect-page.php which both contain redirects to `<doman>/contact-us/`

dhensby: >If we use RedirectorPage::Link() - it'll return the url of the target page

Bummer - that's the worst possible behavior for this module. If the `Link()` function is lying about the page's link then we can't do anything except what's being suggested, I guess.

dhensby (approved):

michalkleiner (approved):
Read the discussion. Unless we change how RedirectorPage works in a BC-breaking way, the solution here is adequate.

PR #174, opened 2023-10-03
MNT Fix unit test in kitchen sink

GuySartorelli:
This test was breaking in kitchen sink because `AbstractQueuedJob` adds specific data when subsites is installed:
https://github.com/symbiote/silverstripe-queuedjobs/blob/d8d4fece6e1e5feec0a3ca6c5c8a0b438a5892c1/src/Services/AbstractQueuedJob.php#L169-L174

## Issue
- https://github.com/silverstripe/.github/issues/125

### DIFF
--- a/tests/php/RunDeleteCacheJobTest.php
+++ b/tests/php/RunDeleteCacheJobTest.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\StaticPublishQueue\Job;
 
 use SilverStripe\Dev\FunctionalTest;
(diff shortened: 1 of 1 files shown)


emteknetnz (approved):

emteknetnz (changes requested):
Should this target 6.0?

GuySartorelli:
Looks like it should, oops. I based it off 6 because that's where the original PR targetted - but it must have been before the stable release.
Now targetting 6.0

emteknetnz (approved):

PR #182, opened 2024-01-11
ENH Add generic types

GuySartorelli:
Requires https://github.com/silverstripe/silverstripe-framework/pull/11108 for these generics to take effect.

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11066

### DIFF
--- a/src/Extension/Publishable/PublishableSiteTree.php
+++ b/src/Extension/Publishable/PublishableSiteTree.php
@@ -19,6 +19,8 @@ use SilverStripe\StaticPublishQueue\Extension\Engine\SiteTreePublishingEngine;
  * an extension via the config system ordering (inject your extension "before" the PublishableSiteTree).
  *
  * @see SiteTreePublishingEngine
+ *
+ * @extends DataExtension<SiteTree>
  */
 class PublishableSiteTree extends DataExtension implements StaticallyPublishable, StaticPublishingTrigger
 {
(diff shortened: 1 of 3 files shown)


src/Extension/Publishable/PublishableSiteTree.php
```diff
@@ -36,7 +38,7 @@ public function getMyVirtualPages()
     }
 
     /**
-     * @return array|SS_List
+     * @return iterable<SiteTree>
```
emteknetnz, line 41: Also update StaticPublishingTrigger return type

src/Extension/Publishable/PublishableSiteTree.php
```diff
@@ -92,7 +94,7 @@ public function objectsToUpdate($context)
     /**
      * This method controls which caches will be purged
      *
-     * @return array|SS_List
+     * @return iterable<SiteTree>
```
emteknetnz, line 97: Also update StaticPublishingTrigger return type

emteknetnz (changes requested):

src/Extension/Publishable/PublishableSiteTree.php
ChloeSartorelli: Good catch. Done

src/Extension/Publishable/PublishableSiteTree.php
ChloeSartorelli: Done

emteknetnz (approved):

PR #201, opened 2024-09-06
API Update API to reflect changes to CLI interaction

GuySartorelli:
Reflects changes in https://github.com/silverstripe/silverstripe-framework/pull/11353

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11341

PR #203, opened 2024-09-10
API Deprecate API that will be removed

GuySartorelli:
See https://github.com/silverstripe/silverstripe-staticpublishqueue/pull/201

## Issues
- https://github.com/silverstripe/silverstripe-framework/issues/11341


emteknetnz (changes requested):
Missing the call to Deprecation::notice() + deprecated annotations

GuySartorelli:
`log()` was the only thing deprecated - the `withNoReplacement` calls all wrap calls to `log()`. Unless I'm not seeing something? If that's the case please comment on the actual code lines that need to be addressed so I can see what you're seeing.

emteknetnz (approved):

PR #208, opened 2024-10-14
ENH Update code to reflect changes in template layer

GuySartorelli:
Relies and reflects on changes in https://github.com/silverstripe/silverstripe-framework/pull/11405

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11322

### DIFF
--- a/tests/php/Publisher/FilesystemPublisherTest.php
+++ b/tests/php/Publisher/FilesystemPublisherTest.php
@@ -140,7 +140,7 @@ class FilesystemPublisherTest extends SapphireTest
 
     public function testMenu2LinkingMode(): void
     {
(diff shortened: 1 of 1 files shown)


tests/php/Publisher/FilesystemPublisherTest.php
```diff
@@ -140,7 +140,7 @@ public function testUrlToPathWithDomainBasedCaching(): void
 
     public function testMenu2LinkingMode(): void
     {
-        SSViewer::set_themes(null);
+        SSViewer::set_themes([]);
```
ChloeSartorelli, line 143: Strict typing added, so we can't pass null anymore.

emteknetnz (approved):

PR #209, opened 2024-10-17
MNT Update unit test

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/320

Location header ended up as `http://example.com/silverstripe.org` - url is missing a scheme and I think the http://example.com came from an alternate_base_url being set elsewhere

### DIFF
--- a/tests/php/Publisher/FilesystemPublisherTest.php
+++ b/tests/php/Publisher/FilesystemPublisherTest.php
@@ -285,7 +285,7 @@ class FilesystemPublisherTest extends SapphireTest
         $redirectorPage = RedirectorPage::create();
         $redirectorPage->URLSegment = 'somewhere-else';
         $redirectorPage->RedirectionType = 'External';
-        $redirectorPage->ExternalURL = 'silverstripe.org';
+        $redirectorPage->ExternalURL = 'https://example.org/path';
         $redirectorPage->write();
         $redirectorPage->publishRecursive();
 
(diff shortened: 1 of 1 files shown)


tests/php/Publisher/FilesystemPublisherTest.php
```diff
@@ -285,7 +285,7 @@ public function testRedirectorPageWhenPHP(): void
         $redirectorPage = RedirectorPage::create();
         $redirectorPage->URLSegment = 'somewhere-else';
         $redirectorPage->RedirectionType = 'External';
-        $redirectorPage->ExternalURL = 'silverstripe.org';
+        $redirectorPage->ExternalURL = 'https://example.com/path';
```
ChloeSartorelli: This test is now inconsistent with `testRedirectorPageWhenHTMLOnly()`. Please update the value in that test as well.

Can you please also confirm that the only problem was the lack of `http(s)://` at the start of the URL? i.e. would `https://silverstripe.org` have worked? If so, this change is fine. Though maybe use `example.org` instead so we have confirmation that it's working and not just using the alternate_base_url.

GuySartorelli (changes requested):

tests/php/Publisher/FilesystemPublisherTest.php
emteknetnz: > Can you please also confirm that the only problem was the lack of http(s):// at the start of the URL?

Yes it was the lack of a scheme, also works with http://

Have changed to https://example.org/path

GuySartorelli (approved):
LGTM

Issue #220, opened 2025-04-03
`https` not used even when `SS_BASE_URL` has been set with the `https` protocol


GuySartorelli:
The code used to check for `SS_STATIC_FORCE_SSL` is broken. This results in always using an `http` URL for all cached content unless you explicitly set `SS_STATIC_FORCE_SSL=true`.

This is because it's checking `is_null(Environment::getEnv('SS_STATIC_FORCE_SSL'))` in `Publisher::generatePageResponse()` - but:
1. Just because I'm not forcing doesn't mean I shouldn't be _allowing_ https
2. `Environment::getEnv()` returns `false` (not `null`) if the environment variable wasn't set

See https://github.com/silverstripe/silverstripe-staticpublishqueue/pull/105 where `SS_STATIC_FORCE_SSL` was introduced - it was only meant to allow _forcing_ https (hence the name), not disable https when not set.

## PRs
- https://github.com/silverstripe/silverstripe-staticpublishqueue/pull/221

PR #221, opened 2025-04-03
FIX Allow https if the URL to cache has an https protocol

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-staticpublishqueue/issues/220

### DIFF
--- a/src/Publisher.php
+++ b/src/Publisher.php
@@ -89,7 +89,7 @@ abstract class Publisher implements StaticPublisher
         }
         try {
             $ssl = Environment::getEnv('SS_STATIC_FORCE_SSL');
-            if (is_null($ssl)) {
+            if (!$ssl) {
                 $ssl = $urlParts['scheme'] == 'https' ? true : false;
             }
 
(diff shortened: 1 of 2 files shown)


src/Publisher.php
```diff
@@ -89,7 +89,7 @@ public function generatePageResponse($url)
         }
         try {
             $ssl = Environment::getEnv('SS_STATIC_FORCE_SSL');
-            if (is_null($ssl)) {
+            if (!$ssl) {
```
emteknetnz, line 92: Will this work with all the variations like 'on' 'off' 'true' 'false' '1' '0'?

Wasn't there some utility that was created to normalise reading environment variables?

emteknetnz (changes requested):

src/Publisher.php
ChloeSartorelli: > Will this work with all the variations like 'on' 'off' 'true' 'false' '1' '0'?

Anything truthy will be truthy, anything falsy will be falsy. Look for example at how we treat `SS_DEPRECATION_ENABLED` or `SS_IGNORE_DOT_ENV` - it's just a truthy check. I don't think we should do anything fancy here.

> Wasn't there some utility that was created to normalise reading environment variables?

We haven't implemented https://github.com/silverstripe/silverstripe-framework/issues/11147 which might be what you're thinking of.

emteknetnz (approved):

PR #233, opened 2026-08-17
MNT Use documentation-lint ^2

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/496


blueo (approved):
checked as a side effect of the yarn 4 upgrade.

